PR #1, opened 2021-12-08
Implement a basic "stack create"

asalkeld:
This downloads templates from github using go-github

### DIFF
--- a/pkg/templates/contents.go
+++ b/pkg/templates/contents.go
@@ -1,0 +1,128 @@
+package templates
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/go-github/v41/github"
+	"golang.org/x/oauth2"
+)
+
+type RepoContent interface {
+	ListSubDirectories(path string) ([]string, error)
+	DownloadDirectoryContents(path string, destDirectory string, force bool) error
+}
+
+type repoContent struct {
+	client *github.Client
+	repo   string
+	owner  string
+}
+
+func NewRepoContent(owner, repo string) RepoContent {
+	ctx := context.Background()
+
+	token := os.Getenv("GITHUB_AUTH_TOKEN")
+	var tc *http.Client
+	if token != "" {
+		ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+		tc = oauth2.NewClient(ctx, ts)
+	} else {
+		fmt.Println("Attempting to download from github without a GITHUB_AUTH_TOKEN")
+	}
+
+	return &repoContent{
+		repo:   repo,
+		owner:  owner,
+		client: github.NewClient(tc),
+	}
+}
+
+func (r *repoContent) ListSubDirectories(path string) ([]string, error) {
+	_, directoryContent, _, err := r.client.Repositories.GetContents(context.Background(), r.owner, r.repo, path, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var subdirs []string
+	for _, c := range directoryContent {
+		switch *c.Type {
+		case "file":
+		case "dir":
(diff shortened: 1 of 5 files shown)


pkg/templates/contents.go
```diff
@@ -0,0 +1,128 @@
...
+func NewRepoContent(owner, repo string) RepoContent {
+	ctx := context.Background()
+
+	token := os.Getenv("GITHUB_AUTH_TOKEN")
+	var tc *http.Client
+	if token != "" {
+		ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+		tc = oauth2.NewClient(ctx, ts)
+	} else {
+		fmt.Println("a GITHUB_AUTH_TOKEN is suggested")
```
asalkeld: need a better message here

pkg/templates/contents.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var subdirs []string
+	for _, c := range directoryContent {
+		switch *c.Type {
+		case "file":
+		case "dir":
+			if !strings.HasPrefix(*c.Path, ".") {
```
asalkeld: need to exclude "e2e-tests"

pkg/templates/contents.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+	return subdirs, nil
+}
+
+func (r *repoContent) DownloadDirectoryContents(path string, destDir string, force bool) error {
+	//	if (fs.existsSync(stackPath) && fs.readdirSync(stackPath).length && !this.force) {
+	//		throw new Error(
+	//			"Stack directory already exists and isn't empty, choose a different name or use the --force flag to create in a non-empty directory.",
+	//		);
```
asalkeld: oops, TODO

pkg/types/stack.go
```diff
@@ -0,0 +1,123 @@
...
+	Name string `yaml:"name"`
+}
+
+type Schedule struct {
+	Expression string `yaml:"expression"`
+
+	// The Topic to be targeted for schedule
+	Target ScheduleTarget `yaml:"target"`
+	Event  ScheduleEvent  `yaml:"event"`
+	//ext?: Ext; TODO(Angus) check
```
tjholm: The intent behind the `ext` fields in the typescript were generic escape hatches that we could add types for in the provider plugins where necessary:

as an example (would allow the AWS deployment to load a "typesafe" configuration for its own extension fields):
```typescript
type AwsSchedule = Schedule<{'x-aws'?: {
  someUniqueAwsProps: string
}}>
```

This would allow us to add cloud specific escape hatches for finetuning that may not be available in other clouds or are challenging to abstract, without generally impacting deployments to other clouds. (could be argued some such config is crucial and other deployments would be broken without it though :P).

The good news is we don't really have this implemented at the moment, so I think leaving it as you have now is perfectly fine. 👍

pkg/types/stack.go
```diff
@@ -0,0 +1,123 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+type Function struct {
+	// The location of the function handler
+	// relative to context
+	Handler string `yaml:"handler"`
+
+	// TODO(Angus) not in the TypeScript
+	Memory int `yaml:"memory,omitempty"`
```
tjholm: We can leave this in 👍  (it should be somewhere in the typescript, but that codebase is a bit of a warren to navigate).

pkg/types/stack.go
```diff
@@ -0,0 +1,123 @@
...
+// A subset of a NitricEvent
+// excluding it's requestId
+// This will be generated based on the scedule
+type ScheduleEvent struct {
+	PayloadType string                 `yaml:"payloadType"`
+	Payload     map[string]interface{} `yaml:"payload,omitempty"`
+}
+
+type ScheduleTarget struct {
+	Type string `yaml:"type"` // TODO(Angus) check type: 'topic'; // ; | "queue"
```
tjholm: I think at the moment we're only supporting topics for schedules at the moment.

PR #7, opened 2022-01-04
implement Endpoints and use the docker client for portablity

asalkeld:


pkg/containerengine/docker.go
```diff
@@ -58,20 +54,24 @@ func newDocker() (ContainerEngine, error) {
...
+		if runtime.GOOS == "linux" {
+			fmt.Println("docker service not running, starting..")
+			cmd = exec.Command("systemctl", "start", "docker")
+			cmd.Stdin = os.Stdin
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			err = cmd.Run()
+			if err != nil {
+				return nil, err
+			}
```
tjholm: Will this require privileged execution in some environments? Might be better to provide an informative error for now.

pkg/containerengine/podman.go
```diff
@@ -0,0 +1,137 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err = cmd.Run()
+	if err != nil || cmd.ProcessState.ExitCode() != 0 {
+		fmt.Println("podman.socket not available, starting..")
+		cmd = exec.Command("sudo", "systemctl", "start", "podman.socket")
+		cmd.Stdin = os.Stdin
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		err = cmd.Run()
```
tjholm: What's the behaviour of this outside of a privileged context? An informative error might be better in this case

tjholm (changes requested):
A few questions from a quick look over

pkg/containerengine/docker.go
asalkeld: it pops up a dialog on my system and asks for sudo password

pkg/containerengine/podman.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for key, val := range buildArgs {
+		args = append(args, "--build-arg="+key+"="+val)
+	}
+	cmd := exec.Command("podman", args...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
```
tjholm: Happy to leave this as is because it's simple. But does the Build method of the docker container engine not work when connecting to podman?

tjholm (commented):
one question around the build method for podman as well.

pkg/containerengine/podman.go
asalkeld: ok

pkg/containerengine/podman.go
asalkeld: turns out the build works fine :-)

tjholm (approved):

PR #8, opened 2022-01-06
AWS pulumi provider

asalkeld:
Deploying nicely, even has consistent taggging so you can make a resource group

### DIFF
--- a/pkg/provider/pulumi/aws/gateway.go
+++ b/pkg/provider/pulumi/aws/gateway.go
@@ -24,7 +24,7 @@ import (
 )
 
 type ApiGatewayArgs struct {
-	ApiFilePath     string
+	OpenAPISpec     *openapi3.T
 	LambdaFunctions map[string]*Lambda
 }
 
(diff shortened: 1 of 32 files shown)


tjholm (approved):
Looks great!

pkg/provider/pulumi/aws/gateway.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func newApiGateway(ctx *pulumi.Context, name string, args *ApiGatewayArgs, opts ...pulumi.ResourceOption) (*ApiGateway, error) {
+	res := &ApiGateway{Name: name}
+	err := ctx.RegisterComponentResource("nitric:api:AwsApiGateway", name, res, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	spec, err := openapi3.NewLoader().LoadFromFile(args.ApiFilePath)
```
tjholm: Forgot about this, this should already be loaded as part of the stack already and shouldn't need to be done again. Currently this field is unexported `stack.apiDocs`. We'll probably want to read from there/export the field or include a method to read it. This will ensure that api deployments can work from exclusively in memory open api specs.

tjholm (changes requested):

jyecusch (approved):

tjholm (approved):

PR #13, opened 2022-01-08
feat: get local ip for docker host

davemooreuws:
Everytime I start my WSL distro, my local IP changes and breaks the build. This fixes the issue by dynamically getting the IP each time.
I am using Windows 10 with WSL2 on a ubuntu distro.
Now this isn't required for native linux, but should still work for it. 

Thoughts please.


tjholm:
@davemooreuws can we test this one windows with the latest build and see if this is resolved?

PR #14, opened 2022-01-10
feat: Add -s <stack> option to "stack describe"

asalkeld:
Rename codeconfig.Stack to codeconfig.CodeConfig

Some other minor things:
- "ctx" is used very widely for context.Context, seeing it in the code for a path is confusing
- allow stderr when running the collection to help see errors
- use the output.Print() function
- add license headers
- store the function name and stack name


pkg/codeconfig/codeconfig.go
```diff
@@ -110,21 +120,32 @@ func (s *Stack) GetApiSpec(api string) (*openapi3.T, error) {
 	return doc, nil
 }
 
-func (s *Stack) String() string {
-	funcStrings := make([]string, 0)
-
-	for _, f := range s.functions {
-		funcStrings = append(funcStrings, f.String())
+func (c *CodeConfig) ToStack() *stack.Stack {
```
asalkeld: @tjholm this is probably wrong, but a start. let me know how I can better do this

tjholm: This is actually pretty close to what I was thinking.

Will add review comments to relevant lines as I go through it.

pkg/codeconfig/codeconfig.go
```diff
@@ -110,21 +123,46 @@ func (s *Stack) GetApiSpec(api string) (*openapi3.T, error) {
...
+	s := &stack.Stack{
+		Name:        path.Base(c.stackPath),
+		Functions:   map[string]stack.Function{},
+		Collections: map[string]interface{}{},
```
tjholm: Might be worth adding placeholders for the missing models here. I know they don't have fields yet, but probably will in the not too distant future

pkg/codeconfig/codeconfig.go
```diff
@@ -110,21 +123,46 @@ func (s *Stack) GetApiSpec(api string) (*openapi3.T, error) {
...
-func NewStack() *Stack {
-	return &Stack{
-		functions: make([]*Function, 0),
-		lock:      sync.RWMutex{},
+	for handler, f := range c.functions {
+		name := strings.Replace(path.Base(handler), path.Ext(handler), "", 1)
+		s.Functions[name] = stack.Function{Handler: handler}
+		for k, v := range f.apis {
+			s.Apis[k] = v.workers[0].String()
+		}
```
tjholm: We'll need a way to pre-populate the api documents as part of the stack. Note the GetApiSpec method on CodeConfig.

pkg/codeconfig/codeconfig.go
```diff
@@ -110,21 +123,46 @@ func (s *Stack) GetApiSpec(api string) (*openapi3.T, error) {
...
+		for k, v := range f.queues {
+			s.Queues[k] = v.String()
+		}
+		for k, v := range f.schedules {
+			s.Schedules[k] = stack.Schedule{
+				Expression: v.String(),
+				Target:     stack.ScheduleTarget{},
+				Event:      stack.ScheduleEvent{},
+			}
```
tjholm, line 326: We should have enough information to provide a target here.

We do need to setup the event as well, but need to do a little bit of design work so we can identify the callback function to route to internally (need just enough information to identify that target internally).

pkg/codeconfig/codeconfig.go
tjholm: One more thought as well is we need to be able to identify if a resource has already been registered by a function, in future we'll need this to implement logic to determine how we should consolidate/merge configuration of these resources.

pkg/codeconfig/codeconfig.go
tjholm: One more thing, it might be good to add an error return on this method as well, to save us refactoring where we're using this method. I know we will eventually return errors in the case where we encounter issues aggregating resources declared by different functions.

pkg/codeconfig/codeconfig.go
asalkeld: @tjholm where is the best place to look for the most complete model definitions? the old cli didn't have anything better.

PR #16, opened 2022-01-11
fix: download stack templates using http and git

davemooreuws:
avoids rate limit issues when using github api and is faster as it downloads the entire sub directory.


pkg/cmd/stack/root.go
```diff
@@ -64,21 +64,38 @@ var stackCreateCmd = &cobra.Command{
...
 		}{}
 
-		rc := templates.NewRepoContent("nitrictech", "stack-templates")
-		dirs, err := rc.ListSubDirectories("")
-		cobra.CheckErr(err)
+		templatesConfig, err := templates.ListTemplates()
+
+		if err != nil {
+			cobra.CheckErr(err)
+		}
```
asalkeld:
```suggestion
		cobra.CheckErr(err)
```

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
-	client *github.Client
-	repo   string
-	owner  string
+type TemplatesConfig struct {
+	Templates []Template
 }
 
-func NewRepoContent(owner, repo string) RepoContent {
-	ctx := context.Background()
+const RAW_GITHUB_URL = "https://raw.githubusercontent.com"
```
asalkeld: can you convert these consts from uppercase to camlecase please
also put them in one block like
```
const (
    rawGitHubURL = "foo"
)
```

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
 
-	return &repoContent{
-		repo:   repo,
-		owner:  owner,
-		client: github.NewClient(tc),
-	}
-}
+const TEMPLATES_REPO_GIT_URL = "github.com/nitrictech/stack-templates.git"
+
+var CONFIG_PATH = path.Join(utils.NitricHome(), "store", "repositories.yml")
```
asalkeld: lowercase camlecase please

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
+
+	return config
 }
 
-func (r *repoContent) DownloadDirectoryContents(path string, destDir string, force bool) error {
-	_, err := os.Stat(destDir)
-	if err == nil && !force {
-		return errors.New("stack directory already exists and isn't empty, choose a different name or use the --force flag to create in a non-empty directory")
-	}
+func GetOfficialStackName(name string) string {
```
asalkeld: Does this need to be public? if a symbol starts with a capital it is public. so consider making this.
Also avoid starting fuctions with "get", see https://go.dev/doc/effective_go#Getters
officialStackName()

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
-}
+	// download file
+	if err := client.Get(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting path %s: %v", client.Src, err)
 
-func (r *repoContent) downloadFile(content *github.RepositoryContent, localPath string, force bool) error {
-	rc, _, err := r.client.Repositories.DownloadContents(context.Background(), r.owner, r.repo, *content.Path, nil)
-	if err != nil {
-		return err
+		return TemplatesConfig{}, err
```
asalkeld: This is a better way of adding info to an error
```
if err := client.Get(); err != nil {
		return TemplatesConfig{}, errors.WithMessagef(err, "Error getting path %s", client.Src)
```

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
-		client: github.NewClient(tc),
-	}
-}
+const TEMPLATES_REPO_GIT_URL = "github.com/nitrictech/stack-templates.git"
+
+var CONFIG_PATH = path.Join(utils.NitricHome(), "store", "repositories.yml")
 
-func (r *repoContent) ListSubDirectories(path string) ([]string, error) {
-	_, directoryContent, _, err := r.client.Repositories.GetContents(context.Background(), r.owner, r.repo, path, nil)
+func ReadTemplatesConfig() TemplatesConfig {
```
asalkeld:
```suggestion
func ReadTemplatesConfig() (*TemplatesConfig,error) {
```

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
+	var config TemplatesConfig
+
+	// Open YAML file
+	file, err := os.Open(CONFIG_PATH)
 	if err != nil {
-		return nil, err
+		log.Println(err.Error())
```
asalkeld:
```suggestion
		return err
```

pkg/templates/contents.go
```diff
@@ -20,125 +20,129 @@ import (
...
-		case "file":
-		case "dir":
-			if !strings.HasPrefix(*c.Path, ".") && strings.Contains(*c.Path, "-stack") {
-				subdirs = append(subdirs, *c.Path)
-			}
+	// Decode YAML file to struct
+	if file != nil {
+		decoder := yaml.NewDecoder(file)
+		if err := decoder.Decode(&config); err != nil {
+			log.Println(err.Error())
```
asalkeld:
```suggestion
			return err
```

asalkeld (changes requested):
nice, just some golang'y suggestions

asalkeld (approved):

PR #23, opened 2022-01-13
Update to allow functions to collect in parallel

tjholm:
Runs function collection in parallel, also doesn't assume port 50051 is available and simply assigns the next available port instead.

Could be cleaned up a bit if we make ErrorsList thread safe. Have simply used sync.Map for now.

### DIFF
--- a/pkg/codeconfig/codeconfig.go
+++ b/pkg/codeconfig/codeconfig.go
@@ -84,7 +84,7 @@ func imageNameFromExt(ext string) string {
 // Collect - Collects information about all functions for a nitric stack
 func (c *codeConfig) Collect() error {
 	wg := sync.WaitGroup{}
-	errs := sync.Map{}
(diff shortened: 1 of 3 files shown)


davemooreuws (approved):

pkg/codeconfig/codeconfig.go
```diff
@@ -83,18 +83,38 @@ func imageNameFromExt(ext string) string {
...
+				errs.Store(rel, err)
+				return
+			}
+		}(f)
 	}
-	return nil
+
+	wg.Wait()
+
+	errList := utils.NewErrorList()
```
asalkeld: couldn't you just add a lock in the errorList.Add()?

tjholm: @asalkeld sure can, wasn't sure if we'd want to make the list itself threadsafe. Will update this now.

PR #30, opened 2022-01-17
Merge WIP nitric run into develop

tjholm:


pkg/cmd/run/root.go
```diff
@@ -0,0 +1,167 @@
...
+			MinWorkers: 0,
+			MaxWorkers: 100,
+		})
+
+		// Start a new gateway plugin
+		gw, err := run.NewGateway()
+
+		if err != nil {
+			cobra.CheckErr(err)
+		}
```
asalkeld:
```suggestion
		cobra.CheckErr(err)
```
no point wrapping this in an err != nil check as this is partly what CheckErr() does

tjholm: fixed.

pkg/cmd/run/root.go
```diff
@@ -0,0 +1,144 @@
...
+		memerr := make(chan error)
+		go func(errch chan error) {
+			errch <- mem.Start()
+		}(memerr)
+
+		time.Sleep(time.Second * time.Duration(2))
+
+		functions, err := run.FunctionsFromHandlers(ctx, files)
+		cobra.CheckErr(err)
+		if err != nil {
```
davemooreuws: left over if condition, remove

pkg/cmd/stack/root.go
```diff
@@ -94,24 +113,69 @@ var stackCreateCmd = &cobra.Command{
...
 	Args: cobra.MaximumNArgs(2),
 }
 
+var stackDescribeCmd = &cobra.Command{
+	Use:   "describe [handler pattern]",
+	Short: "describe stack dependencies",
+	Long:  `Describes stack dependencies`,
+	Run: func(cmd *cobra.Command, args []string) {
+		stackPath, err := filepath.Abs(stack.StackPath())
+		if err != nil {
```
davemooreuws: remove if statement

davemooreuws (changes requested):

pkg/cmd/stack/root.go
tjholm: fixed.

pkg/cmd/run/root.go
tjholm: fixed.

davemooreuws (approved):

davemooreuws:
Approved, linting needs to be fixed as a final change.

PR #31, opened 2022-01-17
feat: worker pool based event routing

jyecusch:


pkg/cmd/run/root.go
```diff
@@ -85,6 +85,12 @@ var runCmd = &cobra.Command{
 		gw, err := run.NewGateway()
 		cobra.CheckErr(err)
 
+		ev, err := run.NewEvents(pool)
+
+		if err != nil {
```
davemooreuws: doesnt require if statement

davemooreuws (changes requested):

tjholm (approved):
looks good to me apart from Dave's requested change and a copyright header that needs fixing.

pkg/provider/run/events.go
```diff
@@ -0,0 +1,85 @@
...
+			"event": event,
+		},
+	)
+
+	requestId := event.ID
+	// payloadType := event.PayloadType
+	payload := event.Payload
+
+	marshaledPayload, err := json.Marshal(payload)
+
```
asalkeld: can we not leave a space between the point of error creation and it's check?
This just makes it clearer what the check is associated with, and doesn't make the code quite so spread out.
like so :
```
x, err = foo()
if err != nil {
    return err
}
```

pkg/provider/run/events.go
```diff
@@ -0,0 +1,85 @@
...
+	targets := s.pool.GetWorkers(&worker.GetWorkerOptions{
+		Event: evt,
+	})
+
+	fmt.Println(fmt.Sprintf("Publishing event to: %s", targets))
+	for _, target := range targets {
+		err = target.HandleEvent(evt)
+
+		if err != nil {
+			fmt.Println(err)
```
asalkeld: so this is just a "user error"? - so not worth changing the flow of our program..

asalkeld (commented):
just stylistic nits

pkg/cmd/run/root.go
jyecusch: fixed

pkg/provider/run/events.go
jyecusch: fixed, added a comment so that intention isn't lost to the ages :)

davemooreuws (approved):
Looks good just fix the lint

PR #37, opened 2022-01-18
Feat/spinners

davemooreuws:
Note: Need to remove the info logs from take port. Maybe use a new library for it.
Includes a fix for no files found on run.

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -133,8 +133,14 @@ func ensureConfigDefaults() {
 	}
 
 	if needsWrite {
-		fmt.Println("updating configfile to include defaults")
-		cobra.CheckErr(viper.WriteConfig())
+		configSpinner, err := output.Spinner("Updating configfile to include defaults")
+		cobra.CheckErr(err)
+		err = viper.WriteConfig()
+		cobra.CheckErr(err)
+
+		// let spinner render
+		time.Sleep(1 * time.Second)
+		configSpinner.Success("Configfile updated")
 	}
 }
 
--- a/pkg/cmd/run/root.go
+++ b/pkg/cmd/run/root.go
@@ -28,6 +28,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/nitrictech/newcli/pkg/build"
+	"github.com/nitrictech/newcli/pkg/output"
 	"github.com/nitrictech/newcli/pkg/run"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/cmd/root.go
```diff
@@ -143,8 +144,18 @@ func ensureConfigDefaults() {
 	}
 
 	if needsWrite {
-		fmt.Println("updating configfile to include defaults")
+		spinner, err := ui.NewSpinner()
+		cobra.CheckErr(err)
+		err = spinner.Start()
+		cobra.CheckErr(err)
+		spinner.Message("Updating configfile to include defaults")
```
asalkeld: This seems really verbose.
How about Rename spinner to subCommand? You could then also be able to deal with timeouts.
```golang
err = subcommand.Run(func (){
    return viper.WriteConfig()
}, subcommand.Opts{Start:"start message", End:"end message", WithSpinner: true, Timeout: 5*time.Second})
cobra.CheckErr(err)
```

pkg/cmd/run/root.go
```diff
@@ -116,6 +135,10 @@ var runCmd = &cobra.Command{
 
 		time.Sleep(time.Second * time.Duration(2))
 
+		spinner.Message("Starting functions")
+
+		time.Sleep(time.Second * time.Duration(1))
```
tjholm: Are these two sleeps in a row necessary?

davemooreuws: Will look at doing sub commands like the suggestion above. Short answer no 👍

pkg/cmd/run/root.go
```diff
@@ -124,7 +147,11 @@ var runCmd = &cobra.Command{
 			cobra.CheckErr(err)
 		}
 
-		fmt.Println("Local running, use ctrl-C to stop")
+		spinner.StopMessage("All services are ready to go! 🚀")
+		err = spinner.Stop()
+		cobra.CheckErr(err)
+
+		fmt.Println("\nLocal running, use ctrl-C to stop")
```
tjholm: This might be printed a bit pre-emptively. We could wait by checking the active worker count if we want to make sure workers will actually be available. If things take a bit of time to warm up users may think that this message is an error when they can't reach their functions.

davemooreuws: Lets make this a new PR

davemooreuws:
Hoping to get this in (for basic spinner functionality) and then others can use the spinner and clean up some of the extra output not required on nitric run. Also checking for works for the final success.

asalkeld (approved):
I am happy to start with this and iterate on it.

PR #38, opened 2022-01-19
Add unit tests for target, cron and templates

asalkeld:


tjholm (approved):
LGTM Would be good to look at some codecov integration as well

PR #39, opened 2022-01-19
feat(run): Separate the run dir from the stack and send output to logs

asalkeld:
This is probably best reviewing commit by commit.


pkg/utils/paths.go
```diff
@@ -50,3 +51,33 @@ func NitricHome() string {
...
+func NitricRunDir() string {
+	if runtime.GOOS == "linux" {
+		u, err := user.Current()
+		if err != nil {
+			log.Fatal(err)
+		}
+		return path.Join("/run/user/", u.Uid, "nitric")
+	}
+	return path.Join(homeDir(), "run")
+}
```
tjholm, line 65: I think this is fine, a few things to think about.

Will all stack runs now share the same directory? If so this could cause parallel runs to pollute each others temporary resources.

Does this prevent a container running as root from creating individual files with root as the owner? As that appeared to the be original error. (I know shifting it here removes it from build context which is good).

It was useful to have temporary stack resources co-located with a project for troubleshooting/debugging (but this can be overcome with documentation).

Could we potentially run the minio container as the current user (probably would be a massive pain to do, but might be worth trying).

Could we run minio embedded instead of running it in a container?

asalkeld: I could add a stackName to the run path as a quick fix..
part 2: i can define a logDir and direct logs there (could be co-located)

other parts, could be added to backlog to investigate?

tjholm: I think the other stuff can be done later as you say, but using the stack name to separate it out is a good quick fix.

davemooreuws (approved):

PR #41, opened 2022-01-20
feat(output): Make the spinner into a tasklet

asalkeld:
The intention is to make the calling code simpler and
to make it more useful.

- catch signals
- handle timeouts
- sleep as needed
- print errors in red


pkg/tasklet/tasklet.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var defaultSequence = []string{"⠟", "⠯", "⠷", "⠾", "⠽", "⠻"}
+
+type StopOn struct {
+	Signal  chan os.Signal
+	Timeout time.Duration
+}
+
+func Do(startMsg string, fn func() error, endMsg string, stop StopOn) error {
```
tjholm: This is great.

Would it be worth separating out the definition of a tasklet from its execution? In the case where we want to generate then run a group of tasklets in sequence?

E.g. `tasklet.New(fun, ...tasklet.Opt)`

It might also be worth coming up with a design that allows the provider of the function to interact with the spinner as well, to allow their code to provide updates to the spinner message.

pkg/tasklet/tasklet.go
```diff
@@ -0,0 +1,75 @@
...
+	start := time.Now()
+	timeout := make(chan bool, 1)
+	done := make(chan bool, 1)
+	doErr := make(chan error, 1)
+	if stop.Timeout > 2*time.Second {
+		go func() {
+			time.Sleep(stop.Timeout)
+			timeout <- true
+		}()
+	}
```
tjholm: could we use a time.Timer here instead?

tjholm (changes requested):

pkg/tasklet/tasklet.go
asalkeld: let me know if the new design is ok..

tjholm: looks great :+1: really good point for us to iterate on as well.

tjholm (approved):

davemooreuws (approved):
Great work

PR #45, opened 2022-01-24
feat: Make builds an implicit part of the deployment

asalkeld:
Allow deployment commands to get gather the stack definition from code.

BREAKING CHANGES: command structure
- remove the build command
- deplyoments now have the "name" passed as a flag
- deployments now accept the handlerGlob


tjholm (approved):

pkg/build/build.go
```diff
@@ -73,14 +74,24 @@ func Create(s *stack.Stack, t *target.Target) error {
...
+		}
+		lang := rt.String()
+		_, ok := imagesToBuild[lang]
+		if ok {
+			continue
+		}
 
-	for lang, imageTag := range imagesToBuild {
-		f, err := os.CreateTemp(stackPath, fmt.Sprintf("%s.*.dockerfile", lang))
+		f, err := os.CreateTemp(s.Path(), fmt.Sprintf("%s.*.dockerfile", lang))
```
tjholm: The stackPath contains nitric.yaml, so nitric.yaml is being treated as a directory in this case.

tjholm: see stack/options.go:56

tjholm (changes requested):

pkg/cmd/deployment/root.go
```diff
@@ -19,79 +19,118 @@ package deployment
...
+		}
 		cobra.CheckErr(err)
+
 		p, err := provider.NewProvider(s, t)
 		cobra.CheckErr(err)
+
+		err = build.Create(s, t)
+		cobra.CheckErr(err)
+
 		cobra.CheckErr(p.Apply(args[0]))
```
tjholm: should apply the stackName.

tjholm (changes requested):

pkg/codeconfig/codeconfig.go
```diff
@@ -43,48 +44,55 @@ import (
...
 
-	for _, f := range c.files {
+	for _, f := range c.initialStack.Functions {
 		wg.Add(1)
 
 		// run files in parallel
 		go func(file string) {
 			defer wg.Done()
-			rel, err := filepath.Rel(c.stackPath, file)
+			rel, err := filepath.Rel(c.initialStack.Dir, file)
```
tjholm: I think the original glob method we were using was fully qualifying the file paths, this is erroring on attempting to make an unrelated path relative (as it already is relative.)

tjholm (changes requested):

pkg/cmd/deployment/root.go
```diff
@@ -19,79 +19,118 @@ package deployment
...
+
 		p, err := provider.NewProvider(s, t)
 		cobra.CheckErr(err)
-		cobra.CheckErr(p.Delete(args[0]))
+
+		cobra.CheckErr(p.Delete(deploymentName))
 	},
-	Args: cobra.ExactArgs(1),
+	Args: cobra.MinimumNArgs(0),
 }
```
tjholm: Just a thought after using this, we don't really need the handler globs in this case. Just a way to get the stack name, and the name of the deployment right? Doesn't need to be done in this PR and happy to discuss it as well.

asalkeld: done

tjholm (approved):

PR #46, opened 2022-01-24
Fix AWS deployments.

tjholm:
Minor fixes for AWS deployments.

still TODO:
 - Need to do more normalization on AWS resource names e.g. remove characters like '.' from lambda names.

### DIFF
--- a/pkg/provider/pulumi/aws/gateway.go
+++ b/pkg/provider/pulumi/aws/gateway.go
@@ -17,6 +17,8 @@
 package aws
 
 import (
+	"fmt"
+
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/pulumi/pulumi-aws/sdk/v4/go/aws/apigatewayv2"
 	awslambda "github.com/pulumi/pulumi-aws/sdk/v4/go/aws/lambda"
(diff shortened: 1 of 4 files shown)


pkg/provider/pulumi/aws/gateway.go
```diff
@@ -35,29 +37,68 @@ type ApiGateway struct {
...
-	b, err := args.OpenAPISpec.MarshalJSON()
+	doc := pulumi.All(nameArnPairs...).ApplyT(func(pairs []interface{}) (string, error) {
+		naps := make(map[string]string)
+
+		for _, p := range pairs {
+			if pair, ok := p.(nameArnPair); ok {
+				naps[pair.name] = pair.invokeArn
+			} else {
+				// XXX: Should not occur
+				return "", fmt.Errorf("invalid data")
```
asalkeld: could have a more meaningful error here.. 
```suggestion
				return "", fmt.Errorf("invalid data %T %v", p, p)
```

pkg/stack/function_helpers.go
```diff
@@ -22,7 +22,7 @@ import (
 	"path/filepath"
 )
 
-const DefaulMembraneVersion = "v0.12.1-rc.5"
+const DefaultMembraneVersion = "v0.13.0-rc.12"
```
asalkeld, line 25: our "make generate" should automatically update this from our go.sum :thinking:

asalkeld (approved):
well done! minor nits if you have time

pkg/provider/pulumi/aws/gateway.go
tjholm: done

pkg/stack/function_helpers.go
tjholm: Good idea, I've been doing a bit more reading on doing this with go generate. Would the idea in this case be that we could create a seperate gen script for individual packages with a generate directive to create that?

In this specific case would we generate a new file e.g. pkg/stack/const.go with this content but have the generate directive in the source file that depends on this constant?

asalkeld: https://github.com/nitrictech/newcli/pull/48

PR #47, opened 2022-01-24
fix: Update the command help and examples for deployment

asalkeld:


pkg/cmd/deployment/root.go
```diff
@@ -71,9 +72,12 @@ var deploymentApplyCmd = &cobra.Command{
 }
 
 var deploymentDeleteCmd = &cobra.Command{
-	Use:   "delete [handlerGlob]",
+	Use:   "delete",
 	Short: "Delete an application deployment",
 	Long:  `Delete a Nitric application deployment.`,
+	Example: `nitric delete -n prod-aws
```
tjholm: Is there a alias/shortcut setup for this or should this be `nitric deployment delete`

pkg/cmd/deployment/root.go
```diff
@@ -86,24 +90,21 @@ var deploymentDeleteCmd = &cobra.Command{
...
-	Args: cobra.MinimumNArgs(0),
+	Args: cobra.ExactArgs(0),
 }
 
 var deploymentListCmd = &cobra.Command{
-	Use:   "list [handlerGlob]",
+	Use:   "list",
 	Short: "list deployments for a stack",
 	Long:  `Lists Nitric application deployments for a stack.`,
+	Example: `nitric list
```
tjholm, line 109: same question for delete

tjholm (commented):
Not sure if these are typos are actual shortcuts available in the CLI

pkg/cmd/deployment/root.go
asalkeld: the deployment name defaults to "dep", so it's not required

PR #48, opened 2022-01-25
build: Generate the default membrane version from the go.sum

asalkeld:
This is to avoid having to manually change it.


pkg/stack/function_helpers.go
```diff
@@ -17,12 +17,14 @@
...
 import (
+	_ "embed"
 	"fmt"
 	"path"
 	"path/filepath"
 )
 
-const DefaulMembraneVersion = "v0.12.1-rc.5"
+//go:embed membraneversion.txt
+var DefaulMembraneVersion string
```
tjholm: There is a typo fix for this on the aws deployments fixes, so I would rebase on this once it's merged to avoid conflicts:
DefaulMembraneVersion -> DefaultMembraneVersion

tjholm (commented):
just one nitpick, otherwise this is awesome!

pkg/stack/function_helpers.go
asalkeld: my special skill is introducing spelling errors into code bases of the world ;-)

tjholm: my signature on any source file I've written is writing recieved instead of received :P.

tjholm (approved):

davemooreuws (approved):

PR #56, opened 2022-01-31
feat: Code-as-config for golang

asalkeld:


pkg/runtime/generate_test.go
```diff
@@ -106,7 +104,7 @@ CMD ["node", "functions/list.js"]`,
 		},
 		{
 			name:     "java",
-			f:        &stack.Function{Handler: "test.java", ComputeUnit: stack.ComputeUnit{Context: "testdata"}},
+			handler:  "testdata/test.java",
```
tjholm, line 107: Just want to make sure with this one, handler entrypoint for java at the moment are always jars (basically just fatjars, that will probably be pulled from /target/ or /build/ (depending on the build system that is being used.

pkg/runtime/types.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type RuntimeExt string
+
+const (
+	RuntimeTypescript RuntimeExt = "ts"
+	RuntimeJavascript RuntimeExt = "js"
+	RuntimePython     RuntimeExt = "py"
+	RuntimeGolang     RuntimeExt = "go"
+	RuntimeJava       RuntimeExt = "java"
```
tjholm, line 47: As mentioned before for java this will be pointed at a `jar` (too much variability in build and dependency systems to reliably build/run from sources.

tjholm (commented):
looks good, just a few things regarding the java runtime.

pkg/runtime/types.go
asalkeld: OK, I suspect java support will need to be fixed up. This change is just moving the existing stuff around, not functionally changing it.

tjholm: no worries, lets look at this in a separate PR.

pkg/runtime/generate_test.go
tjholm: To address in new PR :).

tjholm (approved):

tjholm (approved):

PR #61, opened 2022-02-04
AWS Policies

tjholm:
Apply policies provided by config as code as part of AWS deployments.

### DIFF
--- a/pkg/codeconfig/codeconfig.go
+++ b/pkg/codeconfig/codeconfig.go
@@ -218,8 +218,15 @@ func (c *codeConfig) apiSpec(api string) (*openapi3.T, error) {
 // collectOne - Collects information about a function for a nitric stack
 // handler - the specific handler for the application
 func (c *codeConfig) collectOne(handler string) error {
-	fun := NewFunction()
-	srv := NewServer(fun)
+	rt, err := runtime.NewRunTimeFromHandler(handler)
+	if err != nil {
+		return errors.WithMessage(err, "error getting the runtime from handler "+handler)
+	}
+
+	name := rt.ContainerName()
+	fun := NewFunction(name)
+
+	srv := NewServer(name, fun)
 	grpcSrv := grpc.NewServer()
 
 	v1.RegisterResourceServiceServer(grpcSrv, srv)
@@ -244,11 +251,6 @@ func (c *codeConfig) collectOne(handler string) error {
 	ce, err := containerengine.Discover()
 	if err != nil {
 		return errors.WithMessage(err, "error discovering container engine")
-	}
-
-	rt, err := runtime.NewRunTimeFromHandler(handler)
-	if err != nil {
-		return errors.WithMessage(err, "error getting the runtime from handler "+handler)
 	}
 
 	opts, err := rt.LaunchOptsForFunctionCollect(c.initialStack.Dir)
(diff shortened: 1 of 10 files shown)


pkg/stack/types.go
```diff
@@ -136,6 +137,12 @@ type Stack struct {
 	Schedules   map[string]Schedule    `yaml:"schedules,omitempty"`
 	ApiDocs     map[string]*openapi3.T `yaml:"-"`
 	Apis        map[string]string      `yaml:"apis,omitempty"`
+	// TODO: Not currently supported by nitric.yaml configuration (but is technically definable using the proto model)
+	// We may want to decouple the definition from contracts at a later stage
+	// but re-using the contract here provides us a serializable entity with no
+	// repetition/redefinition
+	// NOTE: if we want to use the proto definition here we would need support for yaml parsing to use customisable tags
```
tjholm, line 144: Not sure if we would want to docouple this definition from the contract policies. I assumed these would end up being much the same anyway in the nitric.yaml.

pkg/stack/types.go
```diff
@@ -136,6 +137,12 @@ type Stack struct {
...
+	Policies []*v1.PolicyResource `yaml:"-"`
```
tjholm, line 146: Not sure if we would want to docouple this definition from the contract policies. I assumed these would end up being much the same anyway in the nitric.yaml.

pkg/codeconfig/codeconfig.go
```diff
@@ -429,14 +435,14 @@ func (c *codeConfig) ToStack() (*stack.Stack, error) {
...
 		if !ok {
-			f = stack.FunctionFromHandler(handler, s.Dir)
+			fun = stack.FunctionFromHandler(handler, s.Dir)
 		}
-		f.ComputeUnit.Triggers = stack.Triggers{
+		fun.ComputeUnit.Triggers = stack.Triggers{
 			Topics: topicTriggers,
 		}
-		s.Functions[rt.ContainerName()] = f
+		s.Functions[rt.ContainerName()] = fun
```
asalkeld: could use "name" here ^

pkg/provider/pulumi/aws/aws.go
```diff
@@ -209,6 +236,24 @@ func (a *awsProvider) Deploy(ctx *pulumi.Context) error {
...
 	}
 
+	for _, p := range a.s.Policies {
+		policyName, err := policyResourceName(p)
+
+		if err != nil {
+			return err
+		}
+
+		newPolicy(ctx, policyName, &PolicyArgs{
```
asalkeld: _, err = newPolicy(...

pkg/stack/types.go
asalkeld: why not just have the FromOpts() populate this with the defaults (max allowed for each resource)

asalkeld: haha, you have the skeleton below - just needs filling in else loading from file will break

tjholm: Yeah put that in so we could just populate default actions for backwards compatibility (Just need to fill it in as you say :)).

This was more a question of whether we wanted to use our auto generated policy definitions from our proto contracts as part of the yaml definition.

eventually we could end up with something like:

```yaml
policies:
  principals:
     - name: test
       type: function
   resources:
     - name: test-topic
       type: topic
   actions:
      # Will want to use a custom deserializer here so we can use strings
      - 400
```

For users who would prefer to use a configuration file. This should almost work as is if we update the tags, howver I'm not sure if we can customise the tags for the yaml lib we're using to also use json tags (which are included on our proto generated sources).

pkg/codeconfig/codeconfig.go
tjholm: fixed.

pkg/provider/pulumi/aws/aws.go
tjholm: fixed

pkg/stack/types.go
tjholm: @asalkeld This should be sorted now

asalkeld (approved):

PR #62, opened 2022-02-07
feat: Add Azure provider

asalkeld:
Should be good to merge now. Tested by @tjholm and @asalkeld


pkg/provider/pulumi/azure/azure.go
```diff
@@ -0,0 +1,179 @@
...
+func (a *azureProvider) SupportedRegions() []string {
+	return []string{
+		"westcentralus",
+		"eastus",
+		"westeurope",
+		"jioindiawest",
+		"northeurope",
+		"canadacentral",
+		"eastus2",
+	}
```
tjholm: It looks like with container apps, we may be limited to eastus2 for now: https://azure.microsoft.com/en-au/global-infrastructure/services/?products=container-apps

pkg/provider/pulumi/azure/schedule.go
```diff
@@ -0,0 +1,105 @@
...
+						"operationTimeout": "00:15:00",
+					},
+				},
+			},
+			"triggers": map[string]interface{}{
+				"Recurrence": map[string]interface{}{
+					"recurrence": trigObj,
+					"type":       "Recurrence",
+				},
+			},
```
asalkeld: just need to figure out the queue send "action"
https://docs.microsoft.com/en-us/azure/connectors/connectors-create-api-servicebus#add-service-bus-action
I might need to use the UI then see if I can export the json to see the format

tjholm (approved):
Port from original CLI looks good, next we'll need to implement deployment of policies (see AWS provider).

PR #63, opened 2022-02-09
Add goreleaser config for homebrew.

tjholm:


davemooreuws (approved):

.goreleaser.yaml
```diff
@@ -38,3 +38,22 @@ scoop:
...
+      name: nitric-bot
+      email: maintainers@nitric.io
+    folder: Formula
+    homepage: "https://nitric.io/"
+    description: "Nitric CLI"
+    license: "Apache 2.0"
+    dependencies:
+      - name: git
+      - name: pulumi # needed for certain provider deployments
+        type: optional
```
asalkeld, line 58: docker?

tjholm: It's a good question, I asked Jye how he'd installed it and it was via the docker website rather than homebrew, not sure if this would confuse with official/brew installations so left out for now. May add back in as optional later.

PR #65, opened 2022-02-09
feat: Allow configured default globs to be used

asalkeld:
This adds the following to the configuration:
```yaml
runtime:
  go:
    functionglob: functions/*/*.go
  ts:
    functionglob: functions/*.ts
```
Then all runtime globs get used if the user does not supply one.
Let me know if these defaults are OK.

### DIFF
--- a/pkg/stack/options.go
+++ b/pkg/stack/options.go
@@ -17,12 +17,16 @@
 package stack
 
 import (
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/pterm/pterm"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
 	"github.com/nitrictech/cli/pkg/pflagext"
 	"github.com/nitrictech/cli/pkg/runtime"
@@ -43,7 +47,7 @@ func wrapStatError(err error) error {
 	return err
 }
 
-func FromOptions() (*Stack, error) {
+func fromNitricFile() (*Stack, error) {
 	configPath := stackPath
 	ss, err := os.Stat(configPath)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/stack/scan.go
```diff
@@ -0,0 +1,74 @@
...
+			switch filepath.Ext(info.Name()) {
+			case ".go":
+				isMain, err := isMainPackage(path)
+				if err != nil {
+					return err
+				}
+				if isMain {
+					funcs = append(funcs, relPath)
+				}
+			case ".ts":
```
asalkeld: not sure how to figure out if this is a "main" file? Any ideas

tjholm: We'll need to use some kind of convention. A few suggestions have been:

Use a directory convention:
e.g. `./functions/**/*.ts`

Or use a file naming convention:
e.g. `./**/*.nitric.ts`

Not sure what everyone prefers though.

Basically it boils down to, if you don't provide a glob then we'll just use the default glob.

asalkeld: thanks @tjholm , I have updated the PR with a per-runtime config option. This will allow the user to change the default too.

pkg/stack/options.go
```diff
@@ -60,44 +63,62 @@ func FromOptions() (*Stack, error) {
...
-		ComputeUnit: ComputeUnit{Name: rt.ContainerName()},
-		Handler:     h,
+func EnsureRuntimeDefaults() bool {
+	defaults := map[string]map[string]interface{}{
+		"ts": {
+			"functionglob": "functions/*.ts",
+		},
+		"go": {
+			"functionglob": "functions/*/*.go",
+		},
```
tjholm, line 74: this is really good!

I think we'll need to consider allowing `.nitric-config` at the project level as well, as I could see this kind of configuration being specific to a project as well as target configuration within a project.

tjholm (approved):

PR #70, opened 2022-02-15
fix: ensure config directory exists

davemooreuws:
Ensures the .config/nitric directory exists before writing config yaml.

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -137,6 +137,12 @@ func ensureConfigDefaults() {
 		tasklet.MustRun(tasklet.Runner{
 			StartMsg: "Updating configfile to include defaults",
 			Runner: func(_ tasklet.TaskletContext) error {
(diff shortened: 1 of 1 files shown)


pkg/cmd/root.go
```diff
@@ -137,6 +137,10 @@ func ensureConfigDefaults() {
 		tasklet.MustRun(tasklet.Runner{
 			StartMsg: "Updating configfile to include defaults",
 			Runner: func(_ tasklet.TaskletContext) error {
+				// ensure .config/nitric exists
+				err := os.MkdirAll(utils.NitricConfigDir(), os.ModePerm)
+				cobra.CheckErr(err)
```
asalkeld:
```suggestion
				if err!=nil{
				return err
				}
```
just so the error will be printed nicely

asalkeld (approved):

PR #73, opened 2022-02-16
feat(deploy): add progress updates

asalkeld:


pkg/provider/pulumi/generator.go
```diff
@@ -130,9 +151,85 @@ func (a *pulumiDeployment) Delete(name string) error {
...
+		event, ok = <-eventChannel
+		if !ok {
+			return
+		}
+		// this is so the message is to the right of the "Deploying ..."
+		spacer := "                     "
+
+		if event.ResourcePreEvent != nil && event.ResourcePreEvent.Metadata.Op != apitype.OpSame {
+			lastCreating, creatingMsg = stepEventToString("ResourcePreEvent", &event.ResourcePreEvent.Metadata)
+			pterm.Info.Print(spacer + creatingMsg)
```
tjholm: We could avoid the output interleaving (spinner being interrupted by output) by using pterm to print our debug output as well e.g. `pterm.Debug.Print()`. This output style co-operates with the spinner on stdout rather than competing.

tjholm (approved):

PR #74, opened 2022-02-16
make building images optional by checking the hash of the contextDir

asalkeld:


tjholm:
I'm wondering if there are more smarts we can do around this? At the moment we're producing a tar for build context, would it take long to take a hash of build context to see if anything has changed and skip building if an image with an attached matching hash already exists? (Container build systems already do this at a layer level anyway as well).

Also what's the behaviour if we're missing the pre-built image?

tjholm (approved):

PR #79, opened 2022-02-21
Add API listing for config as code

tjholm:


asalkeld (commented):
looks good besides the lint error

pkg/cmd/run/root.go
```diff
@@ -126,15 +132,24 @@ var runCmd = &cobra.Command{
 			Api      string `yaml:"api"`
 			Endpoint string `yaml:"endpoint"`
 		}
-		apis := []apiendpoint{}
+		apis := make(map[string]string)
+
+		<-time.NewTimer(time.Second * 2).C
```
tjholm: Wait here is to ensure workers in the pool have had time to register.

We could extend the process pool in the CLI to add event listens and the Add/Remove worker methods to be more reactive about this, but would need a good way to update the output re-actively as well.

asalkeld: could you add a comment here, saying this?

tjholm: Fixing this at the moment by adding a reactive worker pool to the CLI

tjholm: Waited has been removed and replaced by a reactive worker pool.

pkg/run/run.go
```diff
@@ -132,10 +132,10 @@ func (l *localServices) Start() error {
 	// Create a new Worker Pool
 	// TODO: We may want to override GetWorker on the default ProcessPool
 	// For now we'll use the default and expand from there
```
asalkeld: can you just delete this? no point having the commented  out code.

pkg/cmd/run/root.go
```diff
@@ -120,24 +123,27 @@ var runCmd = &cobra.Command{
...
-			apis = append(apis, apiendpoint{Api: a, Endpoint: fmt.Sprintf("http://127.0.0.1:9001/apis/%s", a)})
-		}
-		if len(apis) == 0 {
-			// if we have a nitric.yaml then ApiDocs will be empty
-			for a := range s.Apis {
-				apis = append(apis, apiendpoint{Api: a, Endpoint: fmt.Sprintf("http://127.0.0.1:9001/apis/%s", a)})
-			}
-		}
-		output.Print(apis)
+		area, _ := pterm.DefaultArea.Start()
```
asalkeld, line 148: we probably need something like
```
if s.LoadedFromFile() {
  //watch the file
} else {
  // do the stuff below
}
```

tjholm: I think with the new cli and our documentation we should deprecate yaml support. At least for now

pkg/run/run.go
tjholm: done

jyecusch (approved):

PR #84, opened 2022-02-23
feat: Implement gcp provider

asalkeld:


pkg/provider/pulumi/gcp/cloudrunner.go
```diff
@@ -0,0 +1,185 @@
...
+	sa, err := serviceaccount.NewAccount(ctx, name+"-acct", &serviceaccount.AccountArgs{
+		AccountId: pulumi.String(utils.StringTrunc(name, 30-5) + "-acct"),
+	}, append(opts, pulumi.Parent(res))...)
+	if err != nil {
+		return nil, errors.WithMessage(err, "function serviceaccount "+name)
+	}
+
+	// Give project editor permissions
+	// FIXME: Trim this down
+	_, err = projects.NewIAMMember(ctx, name+"-editor", &projects.IAMMemberArgs{
```
HomelessDinosaur, line 68: Should this be more granular with:
- roles/storage.objectAdmin
- roles/pubsub.editor
- roles/firestore.serviceAgent
- roles/secretmanager.secretAccessor

This will get replaced when the gcp least privilege implementation is done so it may not matter if it has the editor role for the time being.

HomelessDinosaur (changes requested):

pkg/provider/pulumi/gcp/cloudrunner.go
HomelessDinosaur: Otherwise, looks good :)

asalkeld: Yeah, I agree it is not optimal. But good to do separately.

HomelessDinosaur (approved):

PR #107, opened 2022-02-26
fix: Remove membrane childCommand and update all modules

### LINKED ISSUE #99: Error starting the child process: for membrane start on Windows
After successfully creating the dev image and waiting for services to start this error occurs:
"there was an error starting the child process: exec: "echo": executable file not found in %PATH%"

This is caused by the echo in the start process of the membrane.

Environment:
Windows 10 64bit
CLI nitric_1.0.2-develop.2_Windows_x86_64

asalkeld:
This adds a make target "go-mod-update".

fixes #99


tjholm (changes requested):
@asalkeld in the latest RC version of the membrane we should be able to provide a nil child command, without it defaulting.

davemooreuws (approved):

tjholm (approved):

PR #114, opened 2022-02-28
fix: ensure ts/js runtime paths for windows are correct

### LINKED ISSUE #113: Runtime paths are incorrect on Windows for Typescript and Javascript
ts-node doesn't have the correct path and therefore fails.

davemooreuws:
fixes #113


asalkeld (approved):
@davemooreuws is it worth doing a global search/replace for 
path.Join -> filepath.Join
path.Base -> filepath.Base

PR #129, opened 2022-03-01
fix(defaults): Remove the tasklet from around the config update

### LINKED ISSUE #125: config updated message from all commands
SUCCESS  Configfile updated (0s)
is printed as the first output from every command, it probably shouldn't be updated for all commands or at least not printed.

asalkeld:
The update message is not needed.

fixes #125

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -36,7 +36,6 @@ import (
 	"github.com/nitrictech/cli/pkg/output"
 	"github.com/nitrictech/cli/pkg/stack"
 	"github.com/nitrictech/cli/pkg/target"
-	"github.com/nitrictech/cli/pkg/tasklet"
 	"github.com/nitrictech/cli/pkg/utils"
 )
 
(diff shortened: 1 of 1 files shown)


davemooreuws (approved):

pkg/cmd/root.go
```diff
@@ -153,18 +152,12 @@ func ensureConfigDefaults() {
 	}
 
 	if needsWrite {
```
tjholm, line 154: is needsWrite always true? It seems to always run

asalkeld: I know mine definitely doesn't..

tjholm: Running nitric version on a build from develop
image

tjholm: Running version on the latest dev build:
image

I know removing the tasklet will remove the output, but it does run every time on my machine.

asalkeld:
not relevant any more

PR #153, opened 2022-03-04
AWS Resource Discovery

tjholm:
Add permissions and env variables to support resource discovery in AWS.


asalkeld (approved):
lgtm, just merge after the nitric PR?

PR #157, opened 2022-03-07
fix: Prevent codeAsConfig containers from clashing with running containers

### LINKED ISSUE #149: Running multiple instances of nitric
Can’t run nitric on two different terminals at the same time.

```bash
Local running, use ctrl-C to stop

membrane error, exiting: could not listen on configured service address: listen tcp 0.0.0.0:50051: bind: address already in use
```
 
If the projects have conflicting resource names the error is - 

```bash
ERROR   ContainerCreate: Error response from daemon: Conflict. The container name "/public-api" is already in use by container "b8d736c8a91d4edede1f87658b4bab15a1b08d9a8b4be70d42412fe9219be11a". You have to remove (or rename) that container to be able to reuse that name.
ContainerCreate: Error response from daemon: Conflict. The container name "/event-history" is already in use by container "648e0c68cc09112cd6477d4737f1666b36e9139b226363eecbf34071eb65e63d". You have to remove (or rename) that container to be able to reuse that name. of what the bug is.
```

asalkeld:
- name the containers a bit better to prevent scoping issues
- start the LocalServices earlier so we can check Running() and fail earlier

Note: we can only do one "nitric run" per host. If you have different projects you still can't run it twice...
This issue with this is the port we listen on for the membrane.

### DIFF
--- a/pkg/cmd/run/root.go
+++ b/pkg/cmd/run/root.go
@@ -66,6 +66,12 @@ var runCmd = &cobra.Command{
 		}
 		tasklet.MustRun(codeAsConfig, tasklet.Opts{})
 
+		ls := run.NewLocalServices(s)
+		if ls.Running() {
+			pterm.Error.Println("Only one instance of Nitric can be run locally at a time, please check that you have ended all other instances and try again")
+			os.Exit(2)
+		}
+
 		ce, err := containerengine.Discover()
 		cobra.CheckErr(err)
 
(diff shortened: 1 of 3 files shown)


jyecusch (approved):

jyecusch:
I removed the note about fixing #149. It's definitely a good improvement to the experience, so no need to do more right now. However, we could run the membrane on multiple ports or provide some other sort of isolation in future if there is a strong need for multiple local instances at once.

PR #160, opened 2022-03-07
feat: Add a confirmation to the down command

### LINKED ISSUE #151: Add a confirmation y/n question to the Nitric Down command
When the user performs destructive operations like Nitric down, we should give them that information and a chance to back out by selecting or pressing No.

asalkeld:
closes #151

### DIFF
--- a/pkg/cmd/stack/root.go
+++ b/pkg/cmd/stack/root.go
@@ -19,6 +19,7 @@ package project
 import (
 	"fmt"
 	"log"
+	"os"
 	"path/filepath"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -33,6 +34,10 @@ import (
 	"github.com/nitrictech/cli/pkg/provider/types"
 	"github.com/nitrictech/cli/pkg/stack"
 	"github.com/nitrictech/cli/pkg/tasklet"
+)
+
+var (
+	confirmDown bool
 )
 
 var stackCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


pkg/cmd/stack/root.go
```diff
@@ -213,6 +235,7 @@ func RootCommand() *cobra.Command {
 	cobra.CheckErr(stack.AddOptions(stackUpdateCmd, false))
 
 	stackCmd.AddCommand(stackDeleteCmd)
+	stackDeleteCmd.Flags().BoolVarP(&confirmDown, "yes", "y", false, "confirm the destruction of the stack")
```
HomelessDinosaur, line 238: Would it be more intuitive to have the flag as --force/-f to destroy the stack unprompted?

jyecusch: --force is probably better saved for highly destructive operations like deleting buckets containing files or databases.

jyecusch (approved):

pkg/cmd/stack/root.go
raksiv: Can we make this confirmation a little more verbose - 

"Warning - This operation will destroy your stack, all deployed resources will be removed. Are you sure you want to proceed?"

If possible, a better approach might be to show the things being destroyed

"Warning - the following resources will be destroyed:"
     Type                 Name            Plan
 -   Stack           quickstart-dev  delete
 -   ├─ aws:s3:BucketObject  index.html       delete
 -   └─ aws:s3:Bucket        my-bucket           delete

asalkeld: personally think the above is a bit excessive. pulumi doesn't do that do they? I think in general cli tools don't overly baby the user too much. I'll update the text tho'

asalkeld: raised https://github.com/nitrictech/cli/issues/168
For now i'll update the text

HomelessDinosaur (approved):

PR #173, opened 2022-03-09
ci: build and test on ubuntu, windows and mac

asalkeld:
what do people think about this?


davemooreuws:
Great idea. Looks like we may need to change the way it tests on windows.

davemooreuws (approved):

PR #182, opened 2022-03-10
chore: only show subs,topics,apis resources if they are in use

### LINKED ISSUE #175: Don't show resources in output that aren't used
Don't show resources in the output that aren't used.

The screenshot below shows topics and schedules, however none of those are defined in the stack.

image

raksiv:


pkg/cmd/run/root.go
```diff
@@ -155,12 +155,11 @@ var runCmd = &cobra.Command{
...
+
 			area.Update(
-				stackState.ApiTable(9001),
-				"\n\n",
-				stackState.TopicTable(9001),
-				"\n\n",
-				stackState.SchedulesTable(9001),
+				stackState.ApiTable(9001, ""),
+				stackState.TopicTable(9001, "\n\n"),
+				stackState.SchedulesTable(9001, "\n\n"),
```
tjholm: I see that you've added the newlines to conditionally print each of the tables and their suffixes if they are each empty.

Adding the newline characters to the method here makes it less clear what is actually happening. It isn't clear how it's being used or what is going to be printed.

It might be better to externalise the printing logic and add the tables to a slices of strings instead, and finally print the joining of the tables.

e.g.
```go
tables := []string{}

if len(apis) > 0 {
  tables = append(tables, stackState.ApiTable(9001))
}

...other tables...

area.Update(strings.Join(tables, "\n\n"))
```
> Note I haven't checked any of the above for correctness, just a logic example.

tjholm (changes requested):

pkg/cmd/run/root.go
raksiv: Re-written to externalize '\n' formatting.

asalkeld (approved):

davemooreuws (approved):

pkg/cmd/run/root.go
tjholm: @raksiv you legend! looks awesome!

tjholm (approved):

PR #192, opened 2022-03-21
GCP: Least privilege policy implementation

### LINKED ISSUE #196: Least Priviledge - GCP
Implement least privilege resource permissions so that developers only have the minimal set of permissions required to interact with resources they have defined.

tjholm:
Firestore has no least privileged support at the collection level, so we'll need to document that as a caveat for this provider.


pkg/provider/pulumi/gcp/cloudrunner.go
```diff
@@ -56,40 +55,31 @@ func (g *gcpProvider) newCloudRunner(ctx *pulumi.Context, name string, args *Clo
...
+	// // Give secret accessor permissions
+	// _, err = projects.NewIAMMember(ctx, name+"-secret-access", &projects.IAMMemberArgs{
+	// 	Role: pulumi.String("roles/secretmanager.secretAccessor"),
+	// 	// Get the cloudrun service account email
+	// 	Member:  pulumi.Sprintf("serviceAccount:%s", sa.Email),
+	// 	Project: pulumi.String(args.ProjectId),
+	// }, append(opts, pulumi.Parent(res))...)
+	// if err != nil {
+	// 	return nil, errors.WithMessage(err, "iam member "+name)
+	// }
```
asalkeld: not crazy about the commented out code

asalkeld: not crazy about the commented out code

pkg/provider/pulumi/gcp/policy.go
```diff
@@ -0,0 +1,281 @@
...
+
+			switch resource.Type {
+			case v1.ResourceType_Bucket:
+				b := args.Resources.Buckets[resource.Name]
+
+				storage.NewBucketIAMMember(ctx, memberName, &storage.BucketIAMMemberArgs{
+					Bucket: b.Name,
+					Member: memberId,
+					Role:   rolePolicy.Name,
+				}, pulumi.Parent(res))
```
asalkeld, line 229:
```suggestion
				}, append(opts, pulumi.Parent(res))...)
```
as with the others below

asalkeld, line 229:
```suggestion
				}, append(opts, pulumi.Parent(res))...)
```
as with the others below

pkg/provider/pulumi/gcp/policy.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, principal := range args.Policy.Principals {
+		sa := args.Principals[v1.ResourceType_Function][principal.Name]
+		name := newCustomRoleName(principal.Name)
+
+		for _, resource := range args.Policy.Resources {
```
asalkeld, line 217: should we check args.Policy.Resources for nil?

asalkeld, line 217: should we check args.Policy.Resources for nil?

pkg/provider/pulumi/gcp/policy.go
tjholm: I'm not sure we want to propogate resource options to all child resources, but apply them to the parent resource only.

tjholm: resolving as duplicate comment

pkg/provider/pulumi/gcp/policy.go
tjholm: we should implement in the contracts and use generated validation classes, on the gRPC server when we're collecting resources so we can fail as soon as possible.

tjholm: resolving as duplicate comment.

pkg/provider/pulumi/gcp/cloudrunner.go
tjholm: removed

tjholm: removed

pkg/provider/pulumi/gcp/policy.go
asalkeld: ok, well some of the providers do this. so we should at least be consistent one way or the other.

asalkeld:
@tjholm  needs a rebase, but otherwise lgtm

asalkeld (approved):
lgtm

Issue #187, opened 2022-03-18
The ability to set environment variables for local and deploy


davemooreuws:
We need an easy way to set environment variables for local run as well as deployment.

**Why is this needed? What challenge will it help you solve? Please describe.**
It is required to test locally and to deploy a working application that uses environment variables

**Describe alternatives you've considered**
including a .env file in the build, not great.

asalkeld:
take note of k8s https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

but the config-as-code version of this might be to:

1. access the project definition (perhaps inserted into a collection,secret or bucket)
2. access the stack outputs (perhaps inserted into a collection,secret or bucket)
3. access environment variables or settings in the nitric.yaml (as a secret)

What do you think @tjholm

raksiv:
Proposal A  - 

Add support for ports for each API locally
Add configurable ports, so that the local user can specify what ports they want to run locally (this might cause port conflict errors with locally running applications)

asalkeld:
I quite like docker compose .env handling https://docs.docker.com/compose/environment-variables/

How this would work:
1. for "nitric run", look for .local and set this in the running container
2. for "nitric up", look for .production and insert into the functions
3. if none are found either can fallback to .env

If --env-file is provided then this overrides the dot file

PR #204, opened 2022-03-31
feat: add support for .env files

asalkeld:
How this works:

- for "nitric run", look for files in this order --env-file, .local, .env and set this in the running container
- for "nitric up", look for files in this order --env-file, .production, .env  and insert into the functions

(still needs some testing)

implements #187

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/hashicorp/go-getter v1.5.11
 	github.com/imdario/mergo v0.3.12
 	github.com/jedib0t/go-pretty v4.3.0+incompatible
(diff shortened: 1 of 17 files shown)


tjholm (approved):

davemooreuws:
@asalkeld wondering if the files should be:

- .env.local
- .env.production

To be consistent with other frameworks

asalkeld:
> 

@davemooreuws I have changed it to load files in this order:
1. .env
2. .env.{development,production} # development="nitric run", production="nitric up"
3. any file you pass in via "--env-file" as a flag

each load will overwrite an env of the same name in a previous file.

davemooreuws (approved):

PR #224, opened 2022-04-29
feat: use default go path if env var isn't set

jyecusch:
The default go path is ~/go which can be used if the GOPATH variable isn't set. If the home directory can't be determined, we error since the mount will fail anyway

### DIFF
--- a/pkg/runtime/generate_test.go
+++ b/pkg/runtime/generate_test.go
@@ -18,12 +18,12 @@ package runtime
 
 import (
 	"bytes"
-	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/docker/docker/api/types/mount"
 	"github.com/google/go-cmp/cmp"
+	"github.com/nitrictech/cli/pkg/utils"
 )
 
 func TestGenerate(t *testing.T) {
@@ -245,6 +245,11 @@ RUN go install github.com/asalkeld/CompileDaemon@d4b10de`,
 }
 
 func TestLaunchOptsForFunction(t *testing.T) {
+	goPath, err := utils.GoPath()
+	if err != nil {
+		panic(err)
+	}
+
 	tests := []struct {
 		handler string
 		runCtx  string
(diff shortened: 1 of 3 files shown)


tjholm (approved):

pkg/runtime/golang.go
```diff
@@ -131,20 +131,37 @@ func (t *golang) FunctionDockerfileForCodeAsConfig(w io.Writer) error {
 	return err
 }
```
asalkeld: this should work, also check out: https://stackoverflow.com/questions/32649770/how-to-get-current-gopath-from-code
another of doing it is 
```bash
go env GOPATH
```
that should take into account the default and go1.18 way of doing it..

pkg/runtime/generate_test.go
```diff
@@ -245,6 +245,11 @@ RUN go install github.com/asalkeld/CompileDaemon@d4b10de`,
 }
 
 func TestLaunchOptsForFunction(t *testing.T) {
+	goPath, err := utils.GoPath()
+	if err != nil {
+		panic(err)
```
asalkeld:
```suggestion
		t.Error(err)
```

pkg/runtime/generate_test.go
```diff
@@ -311,6 +316,11 @@ func TestLaunchOptsForFunction(t *testing.T) {
 }
 
 func TestLaunchOptsForFunctionCollect(t *testing.T) {
+	goPath, err := utils.GoPath()
+	if err != nil {
+		panic(err)
```
asalkeld:
```suggestion
		t.Error(err)
```

pkg/runtime/generate_test.go
tjholm: fixed.

pkg/runtime/generate_test.go
tjholm: fixed.

pkg/runtime/golang.go
tjholm: fixed.

PR #225, opened 2022-05-11
feat: add a dockerfile for running in gcp cloud build

asalkeld:


Makefile
```diff
@@ -72,3 +72,14 @@ check: lint test generate_check
...
+.PHONY: dockerimage
+dockerimage: build
+	docker build -t asalkeld/nitric-cli:latest .
+	docker tag asalkeld/nitric-cli:latest asalkeld/nitric-cli:x64
+	docker tag asalkeld/nitric-cli:latest asalkeld/nitric-cli:${SOURCE_GIT_COMMIT}
+	docker push asalkeld/nitric-cli:latest
+	docker push asalkeld/nitric-cli:x64
+	docker push asalkeld/nitric-cli:${SOURCE_GIT_COMMIT}
+	docker manifest create asalkeld/nitric-cli:latest --amend asalkeld/nitric-cli:x64  --amend asalkeld/nitric-cli:${SOURCE_GIT_COMMIT}
+	docker manifest push asalkeld/nitric-cli:latest
```
tjholm: Let's use github actions to publish this, I posted a sample in the last review.

tjholm (changes requested):

Makefile
asalkeld: updated now @tjholm can you ptal?

.github/workflows/publish-image.yaml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
asalkeld, line 20: @tjholm I don't have permission to view/set these. Can you please?

tjholm: @asalkeld they are set at the org level so are already available for this repo.

.github/workflows/publish-image.yaml
```diff
@@ -0,0 +1,29 @@
+name: Release Docker images
+on:
+  push:
+    branches:
+    - 'develop'
```
tjholm, line 5: Should this be separated between develop and production? We can do this later, but something we should be aware of.

Makefile
tjholm: @asalkeld looks good, just one more comment/question on develop v. main publishing.

.github/workflows/publish-image.yaml
asalkeld: Well
- our current 'pro' is running in nitric-dev project - it is "dev"
- if the cli image really was only produced via main, I think it would be too slow getting fixes into prod..

tjholm: @asalkeld we wouldn't just produce the image on main, we would always produce on devdelop/main, but the image tags would indicate the stability/version of the image. E.g. dev -> rc-latest (which we could still use with pro) and main -> latest.

This is a rough example as it duplicates the publish action (normally I would conditionally set the tags instead, but this this way is easier than figuring out the multi-line input for now and demonstrates how it's possible).
```yaml
- name: Push Nitric CLI Release Candidate Dockerhub
  if: github.ref == 'refs/heads/develop'
  uses: docker/build-push-action@v3
  with:
    context: .
    file: ./Dockerfile
    push: true
    tags: |
      nitrictech/cli:rc-latest
      nitrictech/cli:rc-x64
- name: Push Nitric CLI Latest Dockerhub
  if: github.ref == 'refs/heads/main'
  uses: docker/build-push-action@v3
  with:
    context: .
    file: ./Dockerfile
    push: true
    tags: |
      nitrictech/cli:latest
      nitrictech/cli:x64
```
While we don't advertise the existence of the CLI docker image yet it's still public, so if we can tag the images to best reflect the release of the build in the image I think we should.

asalkeld: ok, thanks. I'll do that

tjholm (approved):

PR #228, opened 2022-05-13
GCP: Add API Security Definitions

tjholm:


pkg/codeconfig/function.go
```diff
@@ -79,6 +83,25 @@ func (a *Api) AddWorker(worker *pb.ApiWorker) error {
...
+func (a *Api) AddSecurity(name string, scopes []string) {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	sc := []string{}
+	if scopes != nil {
+		sc = scopes
+	}
+
+	a.security[name] = sc
```
asalkeld:
```suggestion
	if scopes != nil {
  	     a.security[name] = scopes
	}
```

asalkeld (commented):
small nit, but looks good

pkg/codeconfig/function.go
tjholm: should `sc` be `scopes` in this change as well? Otherwise looks good :+1:

asalkeld: yip, sorry. i updated the suggestion..

tjholm: just looking at this again the main difference is that `a.security[name]` will always default to `[]string{}` instead of being nil (scopes cannot be nil according the the spec and this prevent the security rule being removed due to `omitEmpty`).

In this case we would want at least:
```go
a.security[name] = []string{}
if scopes != nil {
  a.security[name] = scopes
}
```

tjholm: Have cleaned this up, with an empty slice default.

jyecusch (approved):

pkg/provider/pulumi/gcp/gateway.go
```diff
@@ -228,6 +308,18 @@ func gcpOperation(op *openapi2.Operation, urls map[string]string) *openapi2.Oper
 		return nil
 	}
 
+	if s, ok := op.Extensions["x-nitric-security"]; ok {
+		secName, isString := s.(string)
+
+		if isString {
```
asalkeld, line 314: minor: if it is not a string? I guess if it is nil, that is fine. but do you want to warn if it isn't?

PR #231, opened 2022-05-19
refactor: use hashicorp/go-multierror

asalkeld:


pkg/codeconfig/codeconfig.go
```diff
@@ -102,21 +104,25 @@ func (c *codeConfig) Collect() error {
 			defer wg.Done()
 			rel, err := fn.RelativeHandlerPath(c.initialProject)
 			if err != nil {
-				errList.Add(err)
+				m.Lock()
+				errList = multierror.Append(errList, err)
+				m.Unlock()
```
tjholm: Could we use the multierror `Group` here instead?

asalkeld: it's func does not take an argument, so it's not loop friendly

tjholm (approved):

PR #233, opened 2022-05-23
Feature/kube to manged environment

HomelessDinosaur:
In relation to:
https://github.com/nitrictech/cli/issues/232

Has been tested and fixes the above issue.


pkg/provider/pulumi/azure/azure.go
```diff
@@ -279,7 +279,7 @@ func (a *azureProvider) Deploy(ctx *pulumi.Context) error {
 		Apps:              apps.Apps,
 	})
 	if err != nil {
-		return errors.WithMessage(err, "subscripitons")
```
asalkeld, line 282: omg, that is a sad bug...

asalkeld (approved):

PR #236, opened 2022-05-25
cpu arch support changes

jyecusch:


pkg/cmd/stack/root.go
```diff
@@ -18,14 +18,14 @@ package project
...
-	"path/filepath"
-
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/joho/godotenv"
 	"github.com/pterm/pterm"
 	"github.com/spf13/cobra"
+	"log"
+	"os"
+	"path/filepath"
+	"runtime"
```
asalkeld, line 24: you probably need to revert these import changes

asalkeld (approved):

tjholm (approved):

PR #238, opened 2022-05-26
Gcp api config replace on update

asalkeld:
If the api gateway config ever changes we have an issue on gcp. see:
https://github.com/nitrictech/pro-backend/runs/6601982424?check_suite_focus=true#step:10:1801

I have updates pulumi modules and set the ReplaceOnChanges option on the resource.

also see: https://github.com/hashicorp/terraform-provider-google/issues/8532


pkg/provider/pulumi/gcp/gateway.go
```diff
@@ -176,7 +176,7 @@ func newApiGateway(ctx *pulumi.Context, name string, args *ApiGatewayArgs, opts
 				GoogleServiceAccount: invoker.Email,
 			},
 		},
-	}, opts...)
+	}, append(opts, pulumi.ReplaceOnChanges([]string{"*"}))...)
```
asalkeld, line 179: this is the main change here ^

HomelessDinosaur (approved):

PR #239, opened 2022-05-26
fix(gcp): remove the ApiConfigId property so it can be auto assigned

asalkeld:
Otherwise when it replaces the resource the id will clash.

Tested, and works!


tjholm (approved):
Lgtm

PR #240, opened 2022-06-02
fix(ts): create the tsconfig.json if it does not exist

asalkeld:


pkg/runtime/typescript.go
```diff
@@ -75,6 +75,9 @@ func (t *typescript) FunctionDockerfile(funcCtxDir, version, provider string, w
 		return err
 	}
 
+	tsconfig := `'{"compilerOptions":{"esModuleInterop":true,"target":"es2015","moduleResolution":"node"}}'`
+	buildstage.Run(dockerfile.RunOptions{Command: []string{"echo", tsconfig, ">", "tsconfig.json"}})
```
tjholm: We probably want to conditionally create this if we can (similar to the way we're using `yarn import`), otherwise we'll clobber an existing file.

asalkeld: good idea, now done..

davemooreuws (approved):

PR #251, opened 2022-06-08
fix for gcp secret permissions/listing

asalkeld:
Different permissions need to be applied to different resources.
    - list permissions should be applied to the Project
    - subscripitons.consume needs to be applied to the subscription (not the topic)

Now includes https://github.com/nitrictech/nitric/pull/290

### DIFF
--- a/pkg/provider/pulumi/gcp/policy.go
+++ b/pkg/provider/pulumi/gcp/policy.go
@@ -38,10 +38,11 @@ type Policy struct {
 }
 
 type StackResources struct {
-	Topics  map[string]*pubsub.Topic
-	Queues  map[string]*pubsub.Topic
-	Buckets map[string]*storage.Bucket
-	Secrets map[string]*secretmanager.Secret
+	Topics        map[string]*pubsub.Topic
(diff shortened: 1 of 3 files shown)


pkg/provider/pulumi/gcp/policy.go
```diff
@@ -56,6 +57,12 @@ type PolicyArgs struct {
 	ProjectID pulumi.StringInput
 }
 
+var gcpListActions []string = []string{
+	"pubsub.topics.list",
+	"pubsub.snapshots.list",
+	"resourcemanager.projects.get",
+	"secretmanager.secrets.list",
+}
```
tjholm, line 65: @asalkeld has this been tested? I'm wondering if there are any extra permissions we might need to getting resource labels?

asalkeld: yip I tested against the test app: https://github.com/asalkeld/test-app
and it passes the tests.. (including the new secrets test)

HomelessDinosaur (approved):

PR #258, opened 2022-06-20
Feat/remaining lock file

### LINKED ISSUE #104: When 'nitric up' is interrupted it causes a lock file to be remaining
ERROR   UpsertStackInlineSource: failed to create stack: exit status 255                                                                                                                                                                    
          code: 255
          stdout: 
          stderr: error: could not create stack: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
            file:///home/angus/.pulumi/locks/nitric-chess-api-z/c0325563-63b5-4f4e-90aa-c3700d8e1632.json: created by angus@fedora (pid 193152) at 2022-02-25T15:35:30+10:00

HomelessDinosaur:
Fixes #104

Has to be configurable via an override flag on the `nitric up` command as a conflict is a valid reason to fail if there is more than 1 person working on a stack concurrently. If we make it the default to cancel the updating stack, the other person will lose their update progress. Thus, I've added a flag `-c` or `--cancel`, to cancel the current update in progress.

### DIFF
--- a/pkg/provider/generator.go
+++ b/pkg/provider/generator.go
@@ -26,21 +26,10 @@ import (
 	"github.com/nitrictech/cli/pkg/utils"
 )
 
-func getProviderOpts[T interface{}](opts []interface{}) *T {
-	for _, o := range opts {
-		t, ok := o.(*T)
-		if ok {
-			return t
-		}
-	}
-	return nil
-}
-
-func NewProvider(p *project.Project, s *stack.Config, envMap map[string]string, opts ...interface{}) (types.Provider, error) {
+func NewProvider(p *project.Project, s *stack.Config, envMap map[string]string, opts *types.ProviderOpts) (types.Provider, error) {
 	switch s.Provider {
 	case stack.Aws, stack.Azure, stack.Digitalocean, stack.Gcp:
-		pulumiOpts := getProviderOpts[pulumi.PulumiOpts](opts)
-		return pulumi.New(p, s, envMap, pulumiOpts)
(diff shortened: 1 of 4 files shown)


pkg/provider/generator.go
```diff
@@ -26,10 +26,21 @@ import (
 	"github.com/nitrictech/cli/pkg/utils"
 )
 
-func NewProvider(p *project.Project, s *stack.Config, envMap map[string]string) (types.Provider, error) {
+func getProviderOpts[T interface{}](opts []interface{}) *T {
+	for _, o := range opts {
+		t, ok := o.(*T)
+		if ok {
+			return t
```
asalkeld: isn't this going to ignore all but the first option?

pkg/provider/pulumi/generator.go
```diff
@@ -38,10 +38,15 @@ import (
 	"github.com/nitrictech/cli/pkg/utils"
 )
 
+type PulumiOpts struct {
```
asalkeld: I think it's a bit confusing calling this "Pulumi"opts as we could have a provider that doesn't use pulumi and then what do we do with it? How about StackOpts / StackActionOpts / StackDeployOpts ?
At first read, i thought it was a pulumi type..

asalkeld: then you can move this type up to the top level ..
I think it's better to have generic opts rather than specialized ones for each provider type..

asalkeld (approved):

PR #261, opened 2022-06-21
feat: prompt for feedback/discord on first new project

### LINKED ISSUE #194: Prompt users to provide feedback
Given we want valuable feedback from users of the Nitric CLI, perhaps we could show a text output after the first use of `nitric new` (or another appropriate command?) which prompts with a message like this:

```
Thanks for trying nitric!

If you have feedback you can raise issues on GitHub https://github.com/nitrictech/nitric or come talk with us directly on Discord https://our.discord.link.example
```

HomelessDinosaur:


pkg/cmd/root.go
```diff
@@ -58,9 +64,82 @@ var rootCmd = &cobra.Command{
...
+		num, err = strconv.Atoi(string(contents))
+		if err != nil {
+			return false
+		}
+	}
+	if num == -1 { // Set as do not prompt
+		return false
+	} else if num > requiredTimes { // We can prompt
+		return true
+	} else {
```
jyecusch: Seems like this prompt will appear endlessly after the 6th usage. It seems like something that should happen once or occasionally, not constantly.

pkg/cmd/root.go
```diff
@@ -58,9 +64,82 @@ var rootCmd = &cobra.Command{
...
+		return false
+	}
+	num := 0
+	if contents != nil {
+		num, err = strconv.Atoi(string(contents))
+		if err != nil {
+			return false
+		}
+	}
+	if num == -1 { // Set as do not prompt
```
jyecusch: This `-1` value is implicitly coupled to the `setFeedbackNum(-1)` call in the prompt function, which means if one changes they both need to change - but that's not obvious without reading all of the code.

The relationship would probably be better if it was explicit - e.g. use a constant or maybe store a JSON/YAML value in the document based on a Struct that contains the counts and user preferences.

pkg/cmd/root.go
jyecusch: e.g. once per day/week

jyecusch (dismissed):

pkg/cmd/root.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/nitrictech/cli/pkg/output"
 )
 
+const feedbackMsg = "Thanks for trying nitric!\nIf you have feedback you can raise issues on GitHub https://github.com/nitrictech/nitric or come talk with us directly on Discord https://discord.gg/EBPCEgG9"
```
tjholm: Not sure how we can sort this out, the current invite link contains your username for discord. This isn't a huge issue I think, but good to see if we could get around this.

davemooreuws: @tjholm @HomelessDinosaur how's this link? https://discord.gg/bvTW7aYg6B

davemooreuws: @tjholm maybe we just make a nitric user in discord for this purpose.

tjholm: Created a new invite link with my current user `tjh`: https://discord.com/invite/Webemece5C

tjholm: Will look at server boosting later to get a custom invite link going.

HomelessDinosaur: Change to https://discord.com/invite/Webemece5C, as posted on twitter

asalkeld (approved):

jyecusch (approved):

PR #271, opened 2022-07-13
Don't block on subscription handlers.

tjholm:
Handle subscriptions in parallel to avoid subscribers potentially blocking each other.


asalkeld (approved):
nice!

PR #295, opened 2022-08-05
feat: improve the build time by placing the build inline in pulumi

asalkeld:
part of: https://github.com/nitrictech/pro-backend/issues/166

Currently the images are built first with local names (not including any repo as they are not know)
This means that when running "up" from a new machine the images will not be present. and nothing
will be cached. This PR moves the image building directly into pulumi so that the images are pulled
by pulumi before building.


davemooreuws (approved):
LGTM

PR #319, opened 2022-08-18
Default pulumi passphrase file and set if not already configured.

tjholm:


tjholm:
worth noting that in CI use cases this environment variable should be set, otherwise could lead to unexpected behaviour (getting locked out of deployed stack as passphrase will continue to change).

asalkeld:
just needs a 
```
make fmt
```
@tjholm in CI this is already set

jyecusch (approved):

PR #327, opened 2022-08-31
build CLI binary as multi stage build, instead of copying from context.

tjholm:


Dockerfile
```diff
@@ -1,3 +1,15 @@
...
+
+WORKDIR /app/
+
+# Install dependencies
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+RUN make
```
asalkeld:
```suggestion
RUN make build
```
just to not rely on the default target

asalkeld (commented):
small suggestion

pkg/provider/pulumi/generator.go
```diff
@@ -127,9 +125,7 @@ func (p *pulumiDeployment) load(log output.Progress) (*auto.Stack, error) {
 	for _, plug := range p.prov.Plugins() {
 		log.Busyf("Installing Pulumi plugin %s:%s", plug.Name, plug.Version)
 
-		err = retry.Do(func() error {
-			return s.Workspace().InstallPlugin(ctx, plug.Name, plug.Version)
-		}, retry.Attempts(3), retry.Delay(time.Second))
```
asalkeld: merge conflict?

tjholm: looks like it, will drop this change

asalkeld (approved):

PR #328, opened 2022-09-01
feat: add dockerbuildkit pulumi sdk

asalkeld:
Note, one nice thing with this is if you run update and the code has not changed
```
time nitric up -s test4
 SUCCESS  Configuration gathered (1s)                                                                                                       
 Deployed  Stack (15s)                                                                                                                      
┌─────────────────────────────────────────────────────────────────────────┐
| API       | Endpoint                                                    |
| task-list | https://q0re48ycnj.execute-api.ap-southeast-2.amazonaws.com |
└─────────────────────────────────────────────────────────────────────────┘

real	0m23.868s
user	0m18.936s
```


pkg/provider/pulumi/common/build.go
```diff
@@ -29,7 +29,7 @@ import (
 
 func dynamicDockerfile(dir, name string) (*os.File, error) {
 	// create a more stable file name for the hashing
-	return os.CreateTemp(dir, "nitric."+name+".Dockerfile.*")
+	return os.Create(filepath.Join(dir, "nitric."+name+".Dockerfile"))
```
asalkeld: this is one side-effect, it requires that the docker file is in a subdir of the context.. so we can't put it in a temp dir :shrug:

pkg/pulumi-docker-buildkit/sdk/v1.1.17/dockerbuildkit/doc.go
```diff
@@ -0,0 +1,3 @@
+// A Pulumi package for building Docker images with Buildkit.
+//
+package dockerbuildkit
```
tjholm: would it be possible to move these into a separate public repository? Even if we just use the commit hash of the repo as the go dependency, we could set up proper versioning and automated releases later, or raise an issue for support in their repo and link what we've done to work around lack of support so far.

tjholm (changes requested):

pkg/provider/pulumi/common/build.go
tjholm: that's interesting, they are just executing `buildx` using `-f` for the file reference https://github.com/MaterializeInc/pulumi-docker-buildkit/blob/90a12e5f63871beb10ad1e4ffc6b69abb735281b/cmd/pulumi-resource-docker-buildkit/main.go#L268-L288. 

Last time I tested this with buildkit referencing dockerfiles outside of build context worked fine. Might be worth trying to just run a built like this locally on your machine (e.g. using the CLI image but move the dockerfile somewhere outside of context).

Just tested it locally then with the CLI, and the -f reference works outside of context both relative and absolute.

asalkeld: the issue is here:
https://github.com/MaterializeInc/pulumi-docker-buildkit/blob/90a12e5f63871beb10ad1e4ffc6b69abb735281b/cmd/pulumi-resource-docker-buildkit/main.go#L240
https://github.com/MaterializeInc/pulumi-docker-buildkit/blob/90a12e5f63871beb10ad1e4ffc6b69abb735281b/cmd/pulumi-resource-docker-buildkit/main.go#L419
https://github.com/MaterializeInc/pulumi-docker-buildkit/blob/90a12e5f63871beb10ad1e4ffc6b69abb735281b/cmd/pulumi-resource-docker-buildkit/main.go#L373

it joins the contextDir and the dockerfile path

Dockerfile
```diff
@@ -93,8 +69,9 @@ ARG PULUMI_VERSION
 RUN curl -fsSL https://get.pulumi.com/ | bash -s -- --version $PULUMI_VERSION && \
   mv ~/.pulumi/bin/* /usr/bin
 
-RUN pulumi plugin install resource gcp
-RUN pulumi plugin install resource random
+RUN curl -fsSLo /tmp/dch.tgz https://github.com/docker/docker-credential-helpers/releases/download/${DOCKER_PASS_CH}/docker-credential-pass-${DOCKER_PASS_CH}-amd64.tar.gz && \
+    pulumi plugin install resource docker-credential-helpers ${DOCKER_PASS_CH} -f /tmp/dch.tgz && \
+    rm -f /tmp/dch.tgz
```
tjholm: can this be installed this way as a credential helper?

asalkeld: yeah, I'll revert that.. good catch

tjholm (approved):

PR #329, opened 2022-09-02
feat: pass stack config through to pulumi providers

asalkeld:
I used a stack config of :
```yaml
name: staging
provider: gcp
region: australia-southeast1
project: thing
config:
  default:
    memory: 512
  functions/api/main.go:
    timeout: 30
```

### DIFF
--- a/pkg/project/options.go
+++ b/pkg/project/options.go
@@ -30,44 +30,44 @@ import (
 
 var stackPath string
 
-func FromConfig(p *Config) (*Project, error) {
-	s := New(p)
+func FromConfig(c *Config) (*Project, error) {
+	p := New(c)
 
-	for _, g := range p.Handlers {
-		maybeFile := filepath.Join(s.Dir, g)
+	for _, h := range c.Handlers {
+		maybeFile := filepath.Join(p.Dir, h)
 
 		if _, err := os.Stat(maybeFile); err != nil {
-			fs, err := utils.GlobInDir(stackPath, g)
+			fs, err := utils.GlobInDir(stackPath, h)
 			if err != nil {
 				return nil, err
 			}
 
 			for _, f := range fs {
-				fn, err := FunctionFromHandler(f, s.Dir)
+				fn, err := FunctionFromHandler(f)
 				if err != nil {
 					return nil, err
 				}
 
-				s.Functions[fn.Name] = fn
+				p.Functions[fn.Name] = fn
 			}
 		} else {
-			fn, err := FunctionFromHandler(g, s.Dir)
+			fn, err := FunctionFromHandler(h)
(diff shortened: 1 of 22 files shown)


pkg/project/options.go
```diff
@@ -25,58 +25,81 @@ import (
...
 	if err != nil {
 		return Function{}, err
 	}
 
+	c, ok := s.Config[h]
+	if !ok {
+		return Function{
+			ComputeUnit: ComputeUnit{
+				Name:     rt.ContainerName(),
+				Memory:   128,
```
asalkeld: so one thing is our current gcp has a mem default of 512..

pkg/project/types.go
```diff
@@ -39,14 +39,17 @@ type ComputeUnit struct {
 	// Triggers used to invoke this compute unit, e.g. Topic Subscriptions
 	Triggers Triggers `yaml:"triggers,omitempty"`
 
-	// The memory of the compute instance in MB
+	// The memory of the compute instance in MB: default 128
 	Memory int `yaml:"memory,omitempty"`
 
-	// The minimum number of instances to keep alive
+	// The minimum number of instances to keep alive: default 1
```
asalkeld: should be 0

pkg/project/options.go
```diff
@@ -25,58 +25,81 @@ import (
...
-func FunctionFromHandler(h, stackDir string) (Function, error) {
+func FunctionFromHandler(h string, s *stack.Config) (Function, error) {
 	pterm.Debug.Println("Using function from " + h)
 
 	rt, err := runtime.NewRunTimeFromHandler(h)
 	if err != nil {
 		return Function{}, err
 	}
 
+	c, ok := s.Config[h]
```
asalkeld, line 79: this ^ is not smart and needs the full name of the handler.. could be improved, also i don't have the "default" case yet..

pkg/project/options.go
asalkeld: I have now taken this into account

pkg/stack/types.go
```diff
@@ -25,9 +27,30 @@ const (
...
+	}
+
+	switch provider {
+	case Gcp:
+		ret.Memory = to.IntPtr(512)
+	default:
+		ret.Memory = to.IntPtr(128)
+	}
+
+	return ret
```
tjholm: Would it be possible to shift this config into the providers themselves so they can set their own defaults in the case where its not defined in the configuration?

tjholm (commented):
LGTM just one comment.

@jyecusch was also looking at updating the defaults for AWS to get higher default vCPU allocation.

tjholm (approved):

PR #343, opened 2022-09-15
feat: change stack argument to be optional

### LINKED ISSUE #341: Up command without the stack argument
### Suggestion

Allow `nitric up` to be run without specifying the `-s stackname` argument. If no stacks are found it should prompt to create one, if a single stack is found it should default to it, if multiple stacks are found present a list for the user to pick one.

### Value

It makes the `up` command more convenient, without needing to scan stack files if you can't quite recall the name or only ever work with a single stack.

### Alternatives

An alias - this is ok for a single project but inconvenient when working with many stacks or projects.

HomelessDinosaur:
Closes https://github.com/nitrictech/cli/issues/341

### DIFF
--- a/pkg/stack/options.go
+++ b/pkg/stack/options.go
@@ -51,7 +51,9 @@ func ConfigFromOptions() (*Config, error) {
 	sName := stack // Default to the supplied stack
 	if sName == "" {
 		stacks, err := GetStacks()
-		cobra.CheckErr(err)
+		if err != nil {
+			return nil, err
+		}
 
 		if len(stacks) == 1 {
 			// If there is only one stack use that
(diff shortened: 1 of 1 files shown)


pkg/stack/options.go
```diff
@@ -30,9 +31,42 @@ import (
...
+	return stacks, nil
+}
+
 // Assume the project is in the currentDirectory
 func ConfigFromOptions() (*Config, error) {
-	return configFromFile("nitric-" + stack + ".yaml")
+	sName := stack // Default to the supplied stack
+	if sName == "" {
+		stacks, err := GetStacks()
+		cobra.CheckErr(err)
```
asalkeld:
```
if err!=nil{
return nil, err
}
```
we have a error that is getting returned (same below)

asalkeld (dismissed):

pkg/stack/options.go
HomelessDinosaur: Resolved, copy and paste error 🤦‍♂️

asalkeld (approved):

PR #349, opened 2022-09-16
fix: panic on down/up.

tjholm:


davemooreuws (approved):

pkg/provider/pulumi/aws/aws.go
```diff
@@ -98,12 +98,12 @@ func New(p *project.Project, name string, envMap map[string]string) (common.Pulu
 
 	// Hydrate from file if already exists
 	b, err := os.ReadFile(filepath.Join(p.Dir, "nitric-"+name+".yaml"))
-	if !errors.Is(err, os.ErrNotExist) {
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
asalkeld: good catch, but probably less messy to test errors.Is()

pkg/provider/pulumi/aws/aws.go
```diff
@@ -98,12 +98,12 @@ func New(p *project.Project, name string, envMap map[string]string) (common.Pulu
...
 		return nil, err
 	} else if err == nil {
-		err = yaml.Unmarshal(b, asc)
-		if err != nil {
-			return nil, err
+		marshErr := yaml.Unmarshal(b, asc)
```
asalkeld: i doubt this was causing an issue

pkg/provider/pulumi/aws/aws.go
tjholm: fixed

pkg/provider/pulumi/aws/aws.go
tjholm: fixed.

asalkeld (approved):
nice

PR #354, opened 2022-09-20
feat: remove m1 deploy warning

HomelessDinosaur:


asalkeld:
@HomelessDinosaur do we now support other arches?

HomelessDinosaur:
> @HomelessDinosaur do we now support other arches?

Yea, the change to use buildx means that arm arch is now supported. I have verified with deployments on my M1.

tjholm:
@asalkeld to clarify, using buildkit means that m1 mac devices will auto run workloads using QEMU, when the platform flag is set appropriately, by default this will be `linux/amd64`.

tjholm (approved):

PR #355, opened 2022-09-20
Add nitric start

tjholm:
Adds a services only version of the membrane.

Currently does not offer a solution of loading `.env` for functions (something handled by `nitric run` at the moment). At the moment this would need to be handled by the dev, but we should look at a convenience solution to automatically handle this like a wrapper command in the nitric CLI.

### DIFF
--- a/pkg/cmd/start/root.go
+++ b/pkg/cmd/start/root.go
@@ -38,7 +38,7 @@ import (
 
 var envFile string
 
-var servicesCmd = &cobra.Command{
+var startCmd = &cobra.Command{
 	Use:         "start",
 	Short:       "Run nitric services locally for development and testing",
 	Long:        `Run nitric services locally for development and testing`,
@@ -140,6 +140,5 @@ var servicesCmd = &cobra.Command{
 }
 
 func RootCommand() *cobra.Command {
-	servicesCmd.Flags().StringVarP(&envFile, "env-file", "e", "", "--env-file config/.my-env")
-	return servicesCmd
+	return startCmd
 }
--- a/pkg/cmd/start/root.go
+++ b/pkg/cmd/start/root.go
@@ -36,8 +36,6 @@ import (
 	"github.com/nitrictech/cli/pkg/tasklet"
 )
 
-var envFile string
-
 var startCmd = &cobra.Command{
 	Use:         "start",
 	Short:       "Run nitric services locally for development and testing",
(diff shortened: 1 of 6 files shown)


pkg/cmd/start/root.go
```diff
@@ -0,0 +1,145 @@
...
+
+		_ = area.Stop()
+		// Stop the membrane
+		cobra.CheckErr(ls.Stop())
+	},
+	Args: cobra.ExactArgs(0),
+}
+
+func RootCommand() *cobra.Command {
+	servicesCmd.Flags().StringVarP(&envFile, "env-file", "e", "", "--env-file config/.my-env")
```
asalkeld: why do you have this? it seems to not be supported..

asalkeld (commented):
So one option might be to 
```
nitric start
 # it runs the membrane in the background
 # setup the env variables
 # forks a shell for the user to run the command they want..
go run ./functions/api
```

pkg/cmd/start/root.go
tjholm: need to remove that

tjholm:
> So one option might be to
> .
> ```
> nitric start
>  # it runs the membrane in the background
>  # setup the env variables
>  # forks a shell for the user to run the command they want..
> go run ./functions/api
> ```

@asalkeld this is definetely possible. Wasn't sure how far users would want to go when executing multiple functions in parallel or if a subshell could intefere with existing tooling, but they could wrap more complex setups into an existing script I guess.

so something like `nitric start yarn dev` or as a flag might be a bit cleaner than using args `nitric start -c "yarn dev"`.

asalkeld:
> > So one option might be to
> > .
> > ```
> > nitric start
> >  # it runs the membrane in the background
> >  # setup the env variables
> >  # forks a shell for the user to run the command they want..
> > go run ./functions/api
> > ```
> 
> @asalkeld this is definetely possible. Wasn't sure how far users would want to go when executing multiple functions in parallel or if a subshell could intefere with existing tooling, but they could wrap more complex setups into an existing script I guess.
> 
> so something like `nitric start yarn dev` or as a flag might be a bit cleaner than using args `nitric start -c "yarn dev"`.

I don't think it is a blocker..

asalkeld (approved):

pkg/cmd/start/root.go
tjholm: fixed

PR #359, opened 2022-09-28
Wrap base image builds with the nitric membrane and use base images for nitric run

tjholm:
Benefits:
 - Faster local development (especially on macOS and Windows)
 - Simplifies hot-reloading setup
 - Ability to support custom dockerfiles in future
 - Ability to support remote images in future
 - Removes need to build development images during `nitric up` (this should speed up deploy builds as well)
 - Removes need to separately install node modules during `nitric up` (this should speed up deploy builds as well).
 - `nitric run` provides better indication of final build (base images are the same as production deployments).


pkg/cmd/stack/root.go
```diff
@@ -155,6 +156,16 @@ var stackUpdateCmd = &cobra.Command{
...
+		// build base images on updates
+		createBaseImage := tasklet.Runner{
+			StartMsg: "Building Images",
+			Runner: func(_ output.Progress) error {
+				return build.BuildBaseImages(proj)
+			},
+			StopMsg: "Images Built",
+		}
+		tasklet.MustRun(createBaseImage, tasklet.Opts{})
+
```
asalkeld, line 168: this means we are not going to completely benefit from buildx caching..

asalkeld (approved):

PR #360, opened 2022-09-29
feat: add ecr repository as dependency of images

HomelessDinosaur:


pkg/provider/pulumi/common/image.go
```diff
@@ -72,7 +72,12 @@ func NewImage(ctx *pulumi.Context, name string, args *ImageArgs, opts ...pulumi.
 		},
 	}
 
-	res.DockerImage, err = dockerbuildkit.NewImage(ctx, name+"-image", imageArgs, pulumi.Parent(res))
+	resOpts := []pulumi.ResourceOption{ pulumi.Parent(res) }
+	for _, o := range opts {
+		resOpts = append(resOpts, o)
+	}
+
+	res.DockerImage, err = dockerbuildkit.NewImage(ctx, name+"-image", imageArgs, resOpts...)
```
asalkeld: This is a bit simpler
```suggestion
	res.DockerImage, err = dockerbuildkit.NewImage(ctx, name+"-image", imageArgs, append(opts, pulumi.Parent(res))...)
```

asalkeld (approved):

PR #367, opened 2022-10-06
fix: create stack if none exist when nitric up is run

raksiv:


pkg/cmd/stack/root.go
```diff
@@ -138,7 +115,23 @@ var stackUpdateCmd = &cobra.Command{
 	Example: `nitric stack update -s aws`,
 	Run: func(cmd *cobra.Command, args []string) {
 		s, err := stack.ConfigFromOptions()
-		cobra.CheckErr(err)
+
+		if s == nil {
```
asalkeld:
```suggestion
		if err != nil && strings.Contains(err.Error(), "No nitric stacks found") {
```

pkg/stack/options.go
```diff
@@ -70,7 +70,7 @@ func ConfigFromOptions() (*Config, error) {
 				return nil, err
 			}
 		} else {
-			return nil, errors.New("No nitric stacks found, run `nitric stack new` to create a new stack")
+			return nil, errors.New("No nitric stacks found.")
```
asalkeld: errors are should not be sentence'ified - no capital or full stop

asalkeld (approved):

PR #369, opened 2022-10-06
fix: update error message when pulumi is undetected

raksiv:


pkg/provider/pulumi/generator.go
```diff
@@ -121,15 +121,15 @@ func (p *pulumiDeployment) load(log output.Progress) (*auto.Stack, error) {
 
 	aboutData, err := exec.Command("pulumi", "about", "-j").Output()
 	if err != nil {
```
asalkeld: Let's not assume all errors are because of a missing binary - the user could login incorrectly 
```golang
if strings.Contains(err.Error(), "executable file not found") {
...
```

asalkeld (approved):

PR #370, opened 2022-10-06
fix: update the error message when git is missing.

raksiv:


pkg/templates/contents.go
```diff
@@ -134,7 +134,7 @@ func (d *downloader) repository() error {
 
 	// download file
 	if err := client.Get(); err != nil {
-		return fmt.Errorf("error getting path %s: %w", src, err)
+		return fmt.Errorf("Nitric relies on Git to fetch dependencies. Please refer to the installation instructions - https://nitric.io/docs/installation")
```
asalkeld: I don't think this uses git at all, it does an http download of the git repo. Same below

raksiv: getter is unique in its ability to download both directories and files. It also detects certain source strings to be protocol-specific URLs. For example, "github.com/hashicorp/go-getter" would turn into a Git URL and use the Git protocol.

raksiv: Can you double check? I'm pretty sure this is using git - 

The error the I received was directly as a consequence of removing git from my machine.

asalkeld: I think it would be good to check for the particular error, otherwise you hide all kinds of other errors.

asalkeld: probably this error: https://github.com/hashicorp/go-getter/blob/a2ebce998f8d4105bd4b78d6c99a12803ad97a45/get_git.go#L51

asalkeld (approved):

PR #373, opened 2022-10-10
Add support for delayed SNS messages via AWS step functions.

tjholm:
This deploys to base functionality, but the membrane still needs to be upgraded.

### DIFF
--- a/pkg/provider/pulumi/aws/lambda.go
+++ b/pkg/provider/pulumi/aws/lambda.go
@@ -38,7 +38,7 @@ import (
 type LambdaArgs struct {
 	Client      lambdaiface.LambdaAPI
 	StackName   string
-	Topics      map[string]*sns.Topic
+	Topics      map[string]*Topic
 	DockerImage *common.Image
 	Compute     project.Compute
 	EnvMap      map[string]string
@@ -129,6 +129,8 @@ func newLambda(ctx *pulumi.Context, name string, args *LambdaArgs, opts ...pulum
 		return nil, err
 	}
 
+	// allow lambda to execute step function
+
 	envVars := pulumi.StringMap{
 		"NITRIC_ENVIRONMENT": pulumi.String("cloud"),
 		"NITRIC_STACK":       pulumi.String(args.StackName),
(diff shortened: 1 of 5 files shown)


pkg/utils/contains.go
```diff
@@ -0,0 +1,11 @@
+package utils
+
```
asalkeld: try
```golang
slices.Contains()
```

asalkeld: "golang.org/x/exp/slices"

pkg/provider/pulumi/aws/topic.go
```diff
@@ -0,0 +1,149 @@
...
+
+		return string(def), err
+	}).(pulumi.StringOutput)
+
+	// Deploy a delay manager using AWS step functions
+	// This will enable runtime delaying of event
+	res.Sfn, err = sfn.NewStateMachine(ctx, fmt.Sprintf("%s-delay-ctrl", name), &sfn.StateMachineArgs{
+		RoleArn: sfnRole.Arn,
+		// Apply the same name as the topic to the state machine
+		Tags:       common.Tags(ctx, fmt.Sprintf("%s", name)),
```
asalkeld, line 141:
```suggestion
		Tags:       common.Tags(ctx, fmt.Sprintf("%s-delay-ctrl", name)),
```

pkg/utils/contains.go
tjholm: already deleted.

pkg/provider/pulumi/aws/topic.go
tjholm: The implementation uses the topic name, there is no conflict as we filter by resource type as well. Both the SNS topic and the state machine can share the same name. See lookup code in in the membrane change.

asalkeld: ok, makes senses

pkg/provider/pulumi/aws/lambda.go
```diff
@@ -188,7 +190,7 @@ func newLambda(ctx *pulumi.Context, name string, args *LambdaArgs, opts ...pulum
 			_, err = sns.NewTopicSubscription(ctx, name+t+"Subscription", &sns.TopicSubscriptionArgs{
 				Endpoint: res.Function.Arn,
 				Protocol: pulumi.String("lambda"),
-				Topic:    topic.ID(), // TODO check (was topic.sns)
+				Topic:    topic.Sns.ID(), // TODO check (was topic.sns)
```
asalkeld: do you still need the TODO?

asalkeld (approved):

pkg/provider/pulumi/aws/lambda.go
tjholm: removed.

PR #378, opened 2022-10-13
chore: clean up the dockerfile

asalkeld:
- remove yarn and nodejs
- replace amazon-ecr-credential-helper with docker-credential-pass

this reduces the image size from 1.61G to 1.51G

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,8 +35,6 @@ RUN apt-get update -y && \
   pass \
   unzip && \
   # Get all of the signatures we need all at once.
-  curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key  | apt-key add - && \
-  curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg              | apt-key add - && \
   curl -fsSL https://download.docker.com/linux/debian/gpg          | apt-key add - && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -47,16 +45,10 @@ RUN apt-get update -y && \
...
   rm -rf aws && \
   # Add additional apt repos all at once
-  echo "deb https://deb.nodesource.com/node_14.x $(lsb_release -cs) main"                         | tee /etc/apt/sources.list.d/node.list             && \
-  echo "deb https://dl.yarnpkg.com/debian/ stable main"                                           | tee /etc/apt/sources.list.d/yarn.list             && \
   echo "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable"      | tee /etc/apt/sources.list.d/docker.list           && \
   # Install second wave of dependencies
   apt-get update -y && \
-  apt-get install -y \
-  docker-ce \
-  amazon-ecr-credential-helper \
```
tjholm, line 57: looks good but does the production pro backend still use this?

asalkeld: the change to remove it is merged into main, but it's not running.
prod is waiting for cli to be released. So if this merges, it will be fine.

tjholm (approved):

PR #385, opened 2022-10-21
feat: add support for telemetry

asalkeld:
part of https://github.com/nitrictech/nitric/issues/329


pkg/runtime/wrapper.go
```diff
@@ -39,6 +43,61 @@ CMD [%s]
...
 `
 
+const wrapperDockerFileWithOTel = `
+ARG BASE_IMAGE
+
+FROM ${BASE_IMAGE}
+
+RUN apk update \
+  &&   apk add --no-cache ca-certificates wget shadow \
+  &&   update-ca-certificates
```
tjholm: Is the base image always alpine?

tjholm (changes requested):
Looks pretty good, need more time to go over this, but first thing that pops up is an assumption about base image package managers that could limit future functionality.

pkg/runtime/wrapper.go
asalkeld: I have moved these changes into the base images

pkg/runtime/wrapper.go
```diff
@@ -30,15 +36,72 @@ ARG BASE_IMAGE
...
+FROM ${BASE_IMAGE}
+
+ARG MEMBRANE_URI
+ARG MEMBRANE_VERSION
+
+ENV MEMBRANE_VERSION ${MEMBRANE_VERSION}
+
+RUN apk add --no-cache wget && \
+    wget -q ${MEMBRANE_URI} -O /bin/membrane && \
+    chmod +x-rw /bin/membrane
```
tjholm: Still using `apk` here

tjholm (changes requested):

pkg/runtime/wrapper.go
asalkeld: sorted

tjholm (commented):
@asalkeld are the membrane changes required before this gets released? I'm assuming this can be done independently but wanted to check.

asalkeld:
> 

either first should be fine.

tjholm (approved):

PR #387, opened 2022-10-27
feat: allow extra stack config in the config.default section

asalkeld:
part of https://github.com/nitrictech/pro-backend/issues/408


pkg/provider/pulumi/azure/azure.go
```diff
@@ -186,42 +193,63 @@ func (a *azureProvider) Validate() error {
...
+
+	if dc.AdminEmail == nil {
+		dc.AdminEmail = &a.sc.AdminEmail
+	}
+
+	if dc.Org == nil {
+		dc.Org = &a.sc.Org
+	}
+
+	a.sc.Config[stack.DefaultFunctionConfig] = dc
```
tjholm, line 241: @asalkeld just want to confirm if this is a breaking change for existing stack configurations?

asalkeld:
i'll close for now.. not strictly required

PR #397, opened 2022-11-21
Embed minio into local run.

tjholm:
Embeds minio into local run.

This removes docker as a required dependency for local development. (nitric start can run entirely without docker).

Also improves speed of initial start by including all necessary dependencies in the initial binary.


pkg/run/minio.go
```diff
@@ -54,6 +45,7 @@ const (
 
 // Start - Start the local Minio server
 func (m *MinioServer) Start() error {
+
 	runDir, err := filepath.Abs(m.dir)
```
asalkeld: you are creating m.dir but using a hardcoded dir.. you should probably remove one of those

tjholm: fixed. This will currently use a .nitric directory relative to CWD. In most cases users should use `nitric start` command within their project directory (although this is not a requirement).

tjholm: Also removes the mount user permission errors we had with the container version as this runs minio with their account.

pkg/run/minio.go
```diff
@@ -19,47 +19,28 @@ package run
...
 	"github.com/nitrictech/cli/pkg/utils"
 )
 
 type MinioServer struct {
-	dir     string
-	name    string
-	cid     string
-	ce      containerengine.ContainerEngine
-	apiPort int // external API port from the minio container
+	dir         string
```
asalkeld: `dir` is not used, remove it?

tjholm: good point! also removed name.

asalkeld (approved):
looks great, just linting

tjholm:
closing due to license incompatbility with minio

PR #402, opened 2022-11-21
Change to go mods direct

asalkeld:


Makefile
```diff
@@ -71,5 +71,5 @@ check: lint test generate_check
 
 .PHONY: go-mod-update
 go-mod-update:
-	go get -u $$(go run ./hack/allmods)
+	go get -u $$(go run github.com/asalkeld/go-mods-direct)
```
tjholm: can the script remain in this repository, or be a `nitrictech` repository?

tjholm (dismissed):

Makefile
asalkeld: I use it in pro as well, so I made it separate. I could move it, but then pro would break.. it's a small tool

tjholm: They don't have to be moved at the same time, and it doesn't have to be a permanent move either (can still keep yours as an original):

Have forked it from your repo here:
github.com/nitrictech/go-mods-direct

jyecusch (approved):

PR #403, opened 2022-11-21
Embed seaweedfs into local run.

tjholm:
Using seaweedfs over minio due to licensing.


pkg/run/seaweed.go
```diff
@@ -0,0 +1,110 @@
...
+package run
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
+
+	// "github.com/nitrictech/minio/cmd"
```
asalkeld: remove this if you update the PR

pkg/run/seaweed.go
```diff
@@ -0,0 +1,110 @@
...
+	apiPort int // external API port from the seaweedfs s3 API gateway
+}
+
+const (
+	bucketsDir = "./.nitric/buckets"
+	logsDir    = "/.nitric/logs"
+	runPerm    = os.ModePerm // NOTE: octal notation is important here!!!
+)
+
+// Start - Start the local Minio server
```
asalkeld: nit: there is a lot of name confusion between minio and seaweed.. could just s/minio/seaweed?

pkg/run/storage.go
```diff
@@ -37,8 +37,8 @@ type RunStorageService struct {
 
 const (
```
asalkeld: SEAWEED_ ...

tjholm: These are actually no longer used, so will just delete them (the client still provides them but the server doesn't care.

pkg/run/seaweed.go
tjholm: removed

pkg/run/storage.go
tjholm: removed

pkg/run/seaweed.go
tjholm: fixed

asalkeld (approved):

PR #404, opened 2022-11-22
fix: Add missing gcp Labels

asalkeld:


pkg/provider/pulumi/gcp/gateway.go
```diff
@@ -266,6 +270,7 @@ func newApiGateway(ctx *pulumi.Context, name string, args *ApiGatewayArgs, opts
 		DisplayName: pulumi.String(name + "-gateway"),
 		GatewayId:   pulumi.String(name + "-gateway"),
 		ApiConfig:   pulumi.Sprintf("projects/%s/locations/global/apis/%s/configs/%s", args.ProjectId, res.Api.ApiId, config.ApiConfigId),
+		Labels:      common.Tags(ctx, args.StackID, name+"-gateway"),
```
tjholm: We don't necessarily need to split these based on the entity type. In most cases we'll be searching for a specific item in a query anyway. E.g. the method we use to query these is called `ListGateways`, so configs and APIs are already ignored. If we add special naming to these its something we need to know.

This is a recommendation for a specific approach just something we need to think about/document for ouselves

asalkeld: I am just following the convention we have of labeling the x-nitric-name=`<the resource name>`

asalkeld: just changed to using the base name...
Also added an implementation of Details() to code descovery so everyone doesn't have to do this
```go
if os.Getenv("NITRIC_ENVIRONMENT") != "build" {
		deets, err := mainAPI.Details(context.TODO())
		if err != nil {
			return err
		}
	}
```

tjholm (approved):

PR #405, opened 2022-11-23
Add server.Details() implementation for code discovery

asalkeld:
Right now if you call `myApi.Details()` it will fail with not implemented in the config discovery phase.

### DIFF
--- a/pkg/codeconfig/server.go
+++ b/pkg/codeconfig/server.go
@@ -18,6 +18,7 @@ package codeconfig
 
 import (
 	"context"
+	"fmt"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


pkg/codeconfig/server.go
```diff
@@ -88,6 +89,24 @@ func (s *Server) Declare(ctx context.Context, req *pb.ResourceDeclareRequest) (*
...
+			Details: &pb.ResourceDetailsResponse_Api{
+				Api: &pb.ApiResourceDetails{
+					Url: "http://localhost:50051/apis/" + req.Resource.Name,
+				},
+			},
+		}, nil
+	default:
+		return nil, fmt.Errorf("unsupported resource type %s", req.Resource.Type)
+	}
+}
```
tjholm, line 108: As mentioned by @asalkeld this will prevent config collection from failing if this method is included in inline code (not in a runtime context).

example:
```typescript
const publicApi = api('public');
// Will normally fail at config collection
const deets = publicApi.details();
```
> Note in runtimes like cloud run this may also fail to deploy at runtime as the function will exist before the API gateway is deployed.

Providing mock details to avoid this error might also cause issues though and doesn't really help the user with what the issue actually is:

```typescript
const publicApi = api('public');
// Will normally fail at config collection
const deets = publicApi.details();

// user then does something with details outside of runtime
// this will still fail at config collection time
const result = fetch(deets.url);
```

Ignoring the deployment race condition for Cloud Run, we should educate and provide methods that make it clear how to exclude runtime code from being executing at deploy time. This can currently be done with env variable checks, but we should provide more convenience than that.

```typescript
const dbClient =  atRuntime(() => {
  const client = new DbClient(connectionDetails);
  runMigrations(client);
  return client;
});

// the current alternative at the moment is

let dbClient;
if (process.env.NITRIC_ENVIRONMENT !== "build") {
  dbClient = new DbClient(connectionDetails);
  runMigrations(client);
}
```
> This is just an example, any other suggestions/ideas that are better would be great! Note this doesn't fix the race problem in GCP with api deployments, but will be useful for many other use cases.

PR #406, opened 2022-11-25
Shim pulumi using docker if not on path

tjholm:
This needs some extensive testing on multiple platforms. And fresh installs

Still need to write a shim for windows (currently this will only work on linux/macos).

We should probably provide warnings for the login local stuff as well. And for more serious usage recommend installing pulumi locally.

### DIFF
--- a/pkg/provider/pulumi/pulumi-shim.sh
+++ b/pkg/provider/pulumi/pulumi-shim.sh
@@ -1,17 +1,25 @@
 #!/bin/bash
 
+# This script runs a pulumi as a container compatible with
+# the pulumi automation API.
+
+# This script should only be run in the absence of a local pulumi installation
+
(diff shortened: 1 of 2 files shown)


pkg/provider/pulumi/pulumi-shim.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# This script should only be run in the absence of a local pulumi installation
+
+cmd="$@";
+
+docker run \
+    --rm \
+    --network="host" \
+    -v $HOME/.pulumi:/root/.pulumi \
+    -v $HOME:$HOME \
```
tjholm, line 14: TODO: Need to limit this to necessary directories

tjholm:
closing in favour of assisted install and improved guidance.

PR #412, opened 2022-11-28
Make API and trigger ports dynamic

tjholm:
Example output for feedback:
image

Because we're supporting hot-reloading and the number of workers can change at any time, supporting alphabetical ordering of ports its a bit trickier (but not impossible).

At the moment all ports will be consistently held under their names for the lifetime of a single `nitric start`.

### DIFF
--- a/pkg/cmd/start/root.go
+++ b/pkg/cmd/start/root.go
@@ -103,7 +103,13 @@ var startCmd = &cobra.Command{
 			defer lck.Unlock()
 			// area.Clear()
 
-			stackState.UpdateFromWorkerEvent(we)
+			err := ls.Refresh()
+			if err != nil {
+				cobra.CheckErr(err)
+			}
+
+			area.Update(err)
+			stackState.Update(pool, ls)
 
 			tables := []string{}
 			table, rows := stackState.ApiTable(9001)
--- a/pkg/cmd/start/root.go
+++ b/pkg/cmd/start/root.go
@@ -21,7 +21,6 @@ import (
 	"log"
 	"os"
 	"os/signal"
-	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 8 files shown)


pkg/run/gateway.go
```diff
@@ -119,38 +139,114 @@ func (s *BaseHttpGateway) topic(ctx *fasthttp.RequestCtx) {
...
+
+			if !ok {
+				fhttp := &fasthttp.Server{
+					ReadTimeout:     time.Second * 1,
+					IdleTimeout:     time.Second * 1,
+					CloseOnShutdown: true,
+					Handler:         s.api(api.Api()),
+				}
+
+				lis, err := net.Listen("tcp", "localhost:0")
```
jyecusch: I think we should attempt to open a well known port. If it's unavailable, increment by 1 and try again. Random port assignment is a real pain when working with tools like Postman or Thunder Client.

jyecusch (changes requested):

pkg/run/gateway.go
tjholm: no worries, we can swap that out with something stateful. What port should we start from, and how should this differentiate from static service ports?  (which one should take lowest port priority).

pkg/cmd/start/root.go
```diff
@@ -103,7 +103,13 @@ var startCmd = &cobra.Command{
 			defer lck.Unlock()
 			// area.Clear()
 
-			stackState.UpdateFromWorkerEvent(we)
+			err := ls.Refresh()
+			if err != nil {
+				panic(err)
```
asalkeld: cobra.CheckErr(err)

pkg/run/gateway.go
tjholm: Have given trigger port lowest port priority and used IANA recommended ephemeral port range:
image

pkg/cmd/start/root.go
tjholm: fixed.

pkg/utils/net.go
```diff
@@ -21,6 +21,63 @@ import (
...
+			continue
+		}
+
+		// return the listener
+		return lis, nil
+	}
+
+	return nil, fmt.Errorf("no ports available in range [%d-%d]", options.minPort, options.maxPort)
+}
+
```
tjholm: Could be worth putting this in a separate lib and open sourcing that with a close to drop in replacement to `net.Listen`

e.g. 
```go

lis, err := nextport.Listen("localhost", nextport.Min(9001), nextport.Max(10000))
```

pkg/run/gateway.go
tjholm: If we expose specs on the api ports as well we could hook up import functionality for thunderclient and postman as well:

image

And expose the spec at something like `http://localhost:<port>/spec.json`

asalkeld:
@tjholm the e2e tests are failing as they expect the api on `http://localhost:9001/apis/nitric-testr`
you can override this by setting BASE_URL="what ever you want"

tjholm:
> @tjholm the e2e tests are failing as they expect the api on `http://localhost:9001/apis/nitric-testr` you can override this by setting BASE_URL="what ever you want"

Looks like this needs to be updated, as the trigger URL for topics/schedules is going to be different in this new setup.

asalkeld (approved):

jyecusch (approved):

PR #415, opened 2022-11-30
Add dependency checks and install guides

tjholm:
Requires testing on mac, linux and windows.

Currently automates pulumi install but only links to an install guide for docker.

### DIFF
--- a/pkg/command/dependency.go
+++ b/pkg/command/dependency.go
@@ -1,0 +1,129 @@
+// Copyright Nitric Pty Ltd.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


pkg/command/dependency.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		var installErr error
+		platform := runtime.GOOS
+		switch platform {
+		case "darwin":
+			cmd := exec.Command("sh", "-c", "curl -fsSL https://get.pulumi.com | sh")
+			cmd.Stdout = os.Stdout
+			installErr = cmd.Run()
+		case "linux":
```
asalkeld:
```suggestion
		case "darwin", "linux":
```

asalkeld (approved):

PR #419, opened 2022-12-04
fix(aws): don't create duplicate policies

### LINKED ISSUE #418: AWS: Duplicate policy creation causes conflicts
When creating the same resource with multiple policies, if policies have identical permissions and an identical principal this will lead to conflicts in policy creation under pulumi.

Screenshot_2022-12-03_at_13 14 48

asalkeld:
Note: I don't know the exact use case that caused this.. 

closes #418

### DIFF
--- a/pkg/provider/pulumi/aws/aws.go
+++ b/pkg/provider/pulumi/aws/aws.go
@@ -463,6 +463,8 @@ func (a *awsProvider) Deploy(ctx *pulumi.Context) error {
 		}
 	}
 
+	uniquePolicies := make(map[string]bool)
+
 	for _, p := range a.proj.Policies {
 		if len(p.Actions) == 0 {
 			// note Topic receiving does not require an action.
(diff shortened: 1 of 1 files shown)


pkg/provider/pulumi/aws/aws.go
```diff
@@ -475,6 +477,14 @@ func (a *awsProvider) Deploy(ctx *pulumi.Context) error {
 			return err
 		}
 
+		_, ok := uniquePolicies[policyName]
+		if ok {
+			// no point creating multiple policies with the same content.
```
jyecusch, line 482: We should probably find out what caused this, it could be a mistake in the project that we could highlight as well as adding this protection.

jyecusch (approved):

tjholm (approved):

pkg/provider/pulumi/aws/aws.go
tjholm: I think logging an error would be good if we can highlight this as part of our output logging during deployment.

asalkeld: mmm, let's get this merged and i'll try and do that..

PR #436, opened 2022-12-19
add command to preview deployments

### LINKED ISSUE #432: Preview the resources to the user that would be provisioned
Similar to previewing prior to deleting resources, can we also preview what will be deployed.

Users should be able to abort the operation if they wish.

HomelessDinosaur:
Closes: https://github.com/nitrictech/cli/issues/432

Output Example (changing the name of the API):
Screen Shot 2022-12-19 at 1 51 15 pm


HomelessDinosaur:
Considering removing the "View Live" line as well, as that is pulumi specific and will require them to sign in.

tjholm:
> Considering removing the "View Live" line as well, as that is pulumi specific and will require them to sign in.

@HomelessDinosaur this likely only appears if they're signed into pulumi. It would be interesting the see the result for a local login.

HomelessDinosaur:
> @HomelessDinosaur this likely only appears if they're signed into pulumi. It would be interesting the see the result for a local login.

You're right, tested locally whilst logged out and the link wasn't there.

tjholm (approved):

pkg/provider/pulumi/generator.go
```diff
@@ -288,3 +288,48 @@ func (a *pulumiDeployment) Down(log output.Progress) (*types.Summary, error) {
...
+	for _, line := range lines {
+		if !strings.Contains(line, "pulumi:") { // Remove pulumi specific resources
+			color := colorWhite
+			if strings.Contains(line, " +") { // Creating a resource
+				color = colorGreen
+			} else if strings.Contains(line, " -") { // Deleting a resource
+				color = colorRed
+			} else if strings.Contains(line, " ~") { // Updating a resource
+				color = colorYellow
+			}
```
tjholm, line 324: There may be cases where some lines could contain this substring (I think the leading space will probably occlude most though)

PR #440, opened 2023-01-20
grpc remote deploy

asalkeld:


pkg/run/boltdb.go
```diff
@@ -0,0 +1,520 @@
+// Copyright 2021 Nitric Pty Ltd.
```
tjholm: Can we put the porting of these in a separate PR?

tjholm (changes requested):

pkg/run/boltdb.go
asalkeld: https://github.com/nitrictech/cli/pull/441

go.mod
```diff
@@ -2,10 +2,14 @@ module github.com/nitrictech/cli
 
 go 1.18
 
+replace github.com/nitrictech/nitric/core => ../nitric/core
```
tjholm: A branch reference to to deploy-contract branch would help sort out some of the CI/CD build issues.

currently on commit: `5614c4e59cd9f9b51f30dc6358857d2bdfc92e8e`

pkg/provider/remote/remote.go
```diff
@@ -0,0 +1,352 @@
...
+				}
+			*/
+
+			return res, nil
+		}
+	}
+}
+
+func (p *remoteDeployment) List() (interface{}, error) {
+	return []stackSummary{}, errors.New("not supported for remote deployments")
```
tjholm: Lets remove stackSummary for now and return nil here. If we need to keep stack listing we can raise requirements for the deploy contract against the nitric repo.

pkg/provider/remote/remote.go
```diff
@@ -0,0 +1,352 @@
...
+			for _, topicName := range fv.ComputeUnit.Triggers.Topics {
+				if topicName == target.Name {
+					return fv.Name
+				}
+			}
+		}
+
+		return ""
+	}
+}
```
tjholm: This might seem like a bit of over duplication but I think this logic belongs directly with CodeConfig as another `ToDeployUpRequest` or `ToDeployUpResources`. To save translating legacy concepts like this (if we deprecate/remove the old project definition this leaves more tech debt to address).

For more complex translation like building API documentation we can use shared methods to avoid duplication in these cases.

pkg/provider/remote/remote.go
```diff
@@ -0,0 +1,265 @@
...
+
+type remoteDeployment struct {
+	cfc       types.ConfigFromCode
+	stackName string
+	provider  string
+	conn      *grpc.ClientConn
+	ce        containerengine.ContainerEngine
+	// Container id populated after a call to Start
+	cid string
+}
```
tjholm: Composability will help here for supporting more provider execution types:

```go
type binaryRemoteDeployment struct {
  *remoteDeployment
}

func (p *binaryRemoteDeployment) Up(log output.Progress) (*types.Deployment, error) {
	// TODO: Start binary provider ...	
        dep, err := p.remoteDeployment.Up(log)
	// TODO: Stop Binary provider

        return dep, nil
}
```

Would just need to remove the `conn` from the remote deployment struct, and create/close connections on a per use bases (move these to up/down with a shared utility method for opening the connection.

PR #442, opened 2023-02-03
Add provider command structure

asalkeld:


README.md
```diff
@@ -76,3 +80,4 @@ Documentation for all available commands:
 ## Get Started
 
 Check out the [Nitric docs](https://nitric.io/docs) to see how to get started using Nitric.
+//nitric.io/docs) to see how to get started using Nitric.
```
jyecusch, line 83:
```suggestion
```

pkg/cmd/provider/root.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if version != "" {
+			p.Version = version
+		}
+
+		if downloadURL != "" {
+			p.URL = downloadURL
+
+			if p.Version == "" && strings.Contains(p.URL, "releases/download/v") {
+				verMatch := regexp.MustCompile(`\/releases\/download\/([0-9v.]*)\/*`)
```
jyecusch, line 60: Does it matter that the version component could be blank with this pattern?

e.g. `/releases/download//` this would match but return blank for the selection group. If that matters, I'd change the `*` to `+` on the version match so it always captures a string or errors.

`\/releases\/download\/([0-9v.]+)\/*`

pkg/cmd/provider/root.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		return remote.Install(p)
+	},
+	Args: cobra.ExactArgs(1),
+}
+
+var providerRmCmd = &cobra.Command{
+	Use:     "rm [NAME]",
+	Short:   "remove a previously installed provider",
```
jyecusch, line 76:
```suggestion
	Short:   "Remove a previously installed provider",
```

pkg/provider/remote/install.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+type ProviderInstall struct {
+	Name    string `json:"name"`
+	URL     string `json:"url"`
+	Version string `json:"version"`
+}
+
+func Install(provider *ProviderInstall) error {
+	os.MkdirAll(filepath.Join(utils.NitricHomeDir(), "providers", provider.Name), 0777)
```
jyecusch, line 20: check error

tjholm:
Will revisit this once we need commands for installing custom runtimes, for now auto-download as part of existing provider definition will be supported for all `nitric` providers.

PR #451, opened 2023-02-23
fix: default pulumi to last known working version.

tjholm:


davemooreuws (approved):
Looks good, just the lints. Looked like a timeout, so I re-ran the job

PR #374, opened 2022-10-10
Enable delayed pubsub messages via cloud tasks

tjholm:
Adds base infrastructure. Membrane needs to be upgraded to finalise functionality.

PR #453, opened 2023-02-27
feat: Remote Providers

tjholm:
## ⚠️⚠️⚠️ IMPORTANT TESTING INFO ⚠️⚠️⚠️

When cutting over providers types we need to be sure that persistent objects e.g. collections/secrets/buckets/queues **WILL NOT** be removed when upgrading.

The best way to test this is with a nitric application containing these assets and then performing a deployment with an older version of the CLI, then with the new version re-do the deployment and check if any of these resources have been deleted or any data has been lost, if there are any losses then raise an issue against this thread.

> We want the new providers to be as backward compatible as possible (while still being able to clean up the code/structure).

## 📝 PR Info

Adds the ability to invoke a provider from an installed binary on the users local machine (these will be automatically downloaded in the case of nitric providers).

Depends on release artifacts from:
https://github.com/nitrictech/nitric/pull/374

~~Currently new stack code will generate  yaml files for old provider deployments, users will need to update the provider property to get the new providers:~~

~~e.g. `aws` -> `nitric/aws@0.23.0`~~

> Provider names will automatically be carried over to the new deployment method. The version for these will default to a preset latest known version at compile time. as Using `latest` would cause build inconsistency.

The only provider that contains a breaking stack property change is GCP which users will need to update 

`project` -> `gcp-project-id` (The reason for this is our base providers use `project` as a reserved attribute).

Once providers are released, can look at additional changes to request stack property detail from the providers themselves (once a user selects a provider and version we can download the provider and invoke it to get the template and prompts).


pkg/utils/paths.go
```diff
@@ -53,6 +53,11 @@ func NitricHomeDir() string {
 	return filepath.Join(dirname, ".nitric")
 }
 
+// NitricRunDir returns the directory to place runtime data.
```
HomelessDinosaur, line 56:
```suggestion
// NitricProviderDir returns the directory to place provider deployment binaries.
```

pkg/provider/remote/gcp.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/AlecAivazis/survey/v2"
+	"gopkg.in/yaml.v2"
+)
+
+// TODO: Move this into remote provider logic
+// A provider should be able to produce it's own template for a valid stack specification as stack specifications may change over time
+type gcpProvider struct {
+	*nitricDeployment
+}
```
tjholm, line 32: These are just bridging providers while we shift this functionality into the remote providers themselves, we can probably also inject additional bridging logic into them as well, for migrating breaking changes from older provider definitions.

davemooreuws:
Tasks left to improve the DX:

- [x] Add a prompt for nitric providers if there is no pulumi config detected (give option to auto config for testing)
- [x] Fix provider file mode checks on Windows

davemooreuws (dismissed):

pkg/provider/remote/binary.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/nitrictech/cli/pkg/output"
+	"github.com/nitrictech/cli/pkg/provider/types"
+	"github.com/nitrictech/cli/pkg/utils"
+)
+
+// Provides a binary remote provider type
+type binaryRemoteDeployment struct {
+	providerPath    string
+	autoPulumiLogin bool
```
tjholm: Binary deployments should have no concept of pulumi, I think we should add a facility to add environment variables instead.

even adding a private `env` variable here would be enough. (the nitric provider can rewrite it as they are in the same package scope).

pkg/provider/remote/binary.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (p *binaryRemoteDeployment) startProcess() (*os.Process, error) {
+	cmd := exec.Command(p.providerPath)
+
+	if p.autoPulumiLogin {
+		// Retrieve the current environment variables
+		env := os.Environ()
+		cmd.Env = append(env, "PULUMI_BACKEND_URL=file://~")
+	}
```
tjholm: remove this, logic for appending environment variables can be put in through the nitric provider.

pkg/provider/remote/nitric.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return provFile, nil
+}
+
+func checkPulumiLoginState() (bool, error) {
+	cmd := exec.Command("pulumi", "whoami")
+	_, err := cmd.Output()
+	// if no pulumi login state detected
+	if err != nil {
+		var confirm string = "Yes"
```
tjholm: confirm := "Yes"

tjholm:
Once this is released in develop, need to start testing staging deploy to resolve any issues.

davemooreuws (approved):

PR #458, opened 2023-03-01
fix: Move list permissions to base compute permissions.

tjholm:


davemooreuws (changes requested):
Looks good, just a linting error to fix:
image

HomelessDinosaur (approved):

davemooreuws (approved):

Issue #447, opened 2023-02-22
Design: Dockerfile Support


tjholm:
Support deployment of a target dockerfile within a nitric project.

This will need to include considerations for:

- [ ] Skipping membrane wrapping (useful for single purpose/utility software)
- [ ] Setting build context
- [ ] Setting additional build arguments
- [ ] Setting environment variables
- [ ] Local run environment variables

Acceptance Critera:

- [ ] A design proposal that includes the ability to add a dockerfile as part of a nitric project file as a function (service).

tjholm:
The overall support for this is quite simple, as the CLI already builds docker images as part of the deployment process, the configuration at the moment has some compatability issues.

```yaml
name: my-nitric-project
handlers:
  - "functions/*.ts"
```

Currently we deploy with a simple string for scripts/program targets, but for dockerfiles the number of configurable properties gets a bit more complex.

```yaml
name: my-nitric-project
handlers:
 # Auto detect runtime for simple functions who's dependencies are purely within their languages native dependency system
 - "functions/*.ts"
# Object based definition for dockerfiles
 - dockerfile: docker/my-service.dockerfile
   # Need to be able to give a unique name (we may not be able to infer from the dockerfile as it could be re-used for multiple services
   name: my-function-name
   # Need configurable build context
   context: .
   # Needs configurable builds args
   args:
      MY_BUILD_ARG: SOME_VALUE
   # Need a way to instruct that this artifact may/may not be wrapped prior to deployment
   # Providers are responsible for wrapping, so depending on the runtime target it will also need to implement any special requirements
   nitric: false
```

davemooreuws:
Great example @tjholm !

The only potential changes I would make are moving the docker objects into its own key outside of handlers (maybe dockerHandlers). The reasons for this would be helping parse and validate the yaml in the CLI and separate the concerns of the auto generated handlers.

tjholm:
from discussion:

 - Move containers defintions into a separate `containers` key.
 
 ```yaml
 name: my-nitric-project
handlers:
 # Auto detect runtime for simple functions who's dependencies are purely within their languages native dependency system
 - "functions/*.ts"
containers:
# Object based definition for dockerfiles
 - image: my-locally-built-image # as a URI reference
 - dockerfile: docker/*.dockerfile #required!: glob location of dockerfiles to build
   # Need to be able to give a unique name (we may not be able to infer from the dockerfile as it could be re-used for multiple services
[...]

PR #460, opened 2023-03-08
feat: add ability to use dockerfiles and images as functions

davemooreuws:
Relates to #447 

Things left to do (to be split into another PR):
- nitric wrapper toggle

### DIFF
--- a/pkg/project/config.go
+++ b/pkg/project/config.go
@@ -22,6 +22,9 @@ import (
 	"path/filepath"
 	"reflect"
 
+	"github.com/nitrictech/cli/pkg/utils"
+
+	"github.com/go-playground/validator/v10"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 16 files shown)


pkg/project/options.go
```diff
@@ -60,8 +63,57 @@ func FromConfig(c *Config) (*Project, error) {
 		}
 	}
 
+	for _, c := range c.Containers {
+		if c.Dockerfile != "" {
```
tjholm, line 53: Should we consider config to be invalid in this case and supply a specific error (or collection of errors).

pkg/project/options.go
```diff
@@ -60,8 +63,57 @@ func FromConfig(c *Config) (*Project, error) {
...
+			maybeFile := filepath.Join(p.Dir, c.Dockerfile)
+
+			// Get the absolute path of docker context
+			dockerContext, err := filepath.Abs(filepath.Join(p.Dir, c.Context))
+			if err != nil {
+				return nil, err
+			}
+
+			if _, err := os.Stat(maybeFile); err != nil {
```
tjholm: Is a glob with no glob patterns still a valid glob in this case?

If the file actually exists can we still provide it as a glob and use the same logic branch for both actual files and glob patterns? (where an actual file would just be a single result glob pattern?)

pkg/provider/pulumi/common/image.go
```diff
@@ -75,7 +75,9 @@ func NewImage(ctx *pulumi.Context, name string, args *ImageArgs, opts ...pulumi.
 		return nil, err
 	}
 
-	relDocker, err := filepath.Rel(args.ProjectDir, dockerFilePath)
+	dockerContext := dockerContext(args.ProjectDir, args.Compute)
+
+	relDocker, err := filepath.Rel(dockerContext, dockerFilePath)
```
tjholm: These pulumi builds will be deprecated in the near future, and changes related to their context probably should not be implemented (the resulting docker image doesn't/shouldn't actually copy files as it should only be wrapping an existing docker source image.

tjholm:
LGTM, will look at merging this after:
https://github.com/nitrictech/cli/pull/453

pkg/project/config.go
```diff
@@ -42,16 +45,86 @@ type HandlerConfig struct {
...
+	Handlers   []any          `yaml:"handlers"`
+	Containers []DockerConfig `yaml:"containers" validate:"handlers_containers_required,validate_docker_object,dive"`
+}
+
+type DockerConfig struct {
+	Dockerfile string            `yaml:"dockerfile"`
+	Image      string            `yaml:"image" validate:"omitempty,validate_docker_image"`
+	Context    string            `yaml:"context"`
+	Args       map[string]string `yaml:"args"`
+	Nitric     bool              `yaml:"nitric"`
```
tjholm: Can you add BaseComputeConfig here for configurable typing (will also need to be specific as `inline`.

tjholm:
Have some new thoughts on how we can reshape this.

Instead of implementing a new concept we can probably add packaging level customisation directly to the handlers instead.

e.g.

```yaml
handlers:
  - match: './functions/*.ts'
    type: big-ol-container
    docker:
       file: ./docker/big-ol-container.dockerfile
       args:
           # Adding build args here will allow dockerfile re-use
           SOME_BUILD_ARG: some-custom-value
```

From here we will still use the same build system but we'll substitute our OOTB template with theirs, but will inject the handler references into their dockerfile the same way we do with ours, and document how to build a custom container this way.

davemooreuws:
Converted to draft. Can be revisited later if required.

PR #462, opened 2023-03-22
throw an error when a function name with a special character is used.

### LINKED ISSUE #423: Files with spaces in names cause issue when trying to deploy
Running 'nitric up' on a project which has a file with a space in the name within the configured glob folder (default 'functions') then I get an error 'exit status 1' 

No other logs found.

raksiv:
Closes: https://github.com/nitrictech/cli/issues/423

### DIFF
--- a/pkg/project/options.go
+++ b/pkg/project/options.go
@@ -25,6 +25,8 @@ import (
 
 	"github.com/nitrictech/cli/pkg/runtime"
 	"github.com/nitrictech/cli/pkg/utils"
+
+	"github.com/docker/distribution/reference"
 )
 
 var stackPath string
@@ -67,6 +69,11 @@ func FromConfig(c *Config) (*Project, error) {
 }
 
 func FunctionFromHandler(h string, t string) (Function, error) {
+	_, err := reference.Parse(h)
+	if err != nil {
+		return Function{}, fmt.Errorf("handler filepath \"%s\" is invalid, must be valid ASCII containing lowercase and uppercase letters, digits, underscores, periods and hyphens", h)
+	}
+
 	pterm.Debug.Println("Using function from " + h)
 
 	rt, err := runtime.NewRunTimeFromHandler(h)


pkg/project/options.go
```diff
@@ -68,6 +69,10 @@ func FromConfig(c *Config) (*Project, error) {
 }
 
 func FunctionFromHandler(h string) (Function, error) {
+	if regexp.MustCompile(`[~^\[\]()#$*&!\s]`).MatchString(h) {
```
jyecusch: How did you determine the list of characters to check for, are we sure this is the full list?

raksiv: The error is caused by tag names in docker - "A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and hyphens.".

Tested the ones included in the list caused the issue - it's possible there are more. I will try and find more info or spot-check chars like `@%+=|?<>, to see if they need to be added.

tjholm: does docker already have a regex for this that we can reuse?

davemooreuws: On my local, I was working on the docker image support. I created an image string parser util for docker (including the tag)
So you could re-use some of this to validate if you like, as it uses "github.com/docker/distribution/reference"

```go
package utils

import (
	"bytes"

	"github.com/docker/distribution/reference"
)

func StringTrunc(s string, max int) string {
	if len(s) <= max {
		return s
	}

	return s[:max]
}

func JoinCamelCase(ss []string) string {
	res := ss[0]

	for i := 1; i < len(ss); i++ {
		word := ss[i]
		res += string(bytes.ToUpper([]byte{word[0]}))
		res += word[1:]
	}

	return res
}

type DockerImageMeta struct {
	Name string
	Tag string
}

type ParsedDocker interface {
	String() string
	Name() string
	Tag() string
}

func ParseDockerImage(image string) (DockerImageMeta, error) {
	meta := DockerImageMeta{}

	imageRef, err := reference.Parse(image)
    if err != nil {
        return meta, err
    }

	namedRef, ok := imageRef.(reference.Named)
    if !ok {
		return meta, err
    }

	meta.Name = reference.Path(namedRef)

	taggedRef, ok := imageRef.(reference.Tagged)
    if ok {
        meta.Tag = taggedRef.Tag()
    }

	return meta, nil
}
```

raksiv: using the parser from "github.com/docker/distribution/reference" seems to make the most sense.
I've removed the regexps and used this function - also realigned the error message with the docker docs.

tjholm (approved):

davemooreuws (approved):

davemooreuws (changes requested):
Looks good, just requires a rebase

raksiv:
rebased onto dev

davemooreuws (approved):

PR #470, opened 2023-03-31
add custom deployment provider support

HomelessDinosaur:


pkg/provider/remote/custom.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v2"
+)
+
+// TODO: Move this into remote provider logic
+// A provider should be able to produce it's own template for a valid stack specification as stack specifications may change over time
```
jyecusch, line 28: Can this be done before the PR is merged?

pkg/provider/remote/custom.go
```diff
@@ -0,0 +1,41 @@
...
+type customProvider struct {
+	*nitricDeployment
+}
+
+// FIXME: Prompting to create a new stack state and memory and persisting to a file should be two separte functions
+func (g *customProvider) AskAndSave() error {
```
jyecusch, line 34: This too?

HomelessDinosaur: Not yet, remote providers can't make templates yet cause there's no API for it. Will need to create a template and prompt API that the CLI can understand.

HomelessDinosaur: Same with above ^^

pkg/provider/remote/custom.go
tjholm: This needs some work on the contracts for the remote providers, Probably some kind of service we can add the the proto contracts that allows us to output a template, but would need to be readable by the CLI, for user prompts/token replacement.

davemooreuws (approved):

PR #480, opened 2023-05-01
feat: bucket notifications

HomelessDinosaur:
Relies on contract changes from: https://github.com/nitrictech/nitric/pull/428


pkg/run/storage.go
```diff
@@ -64,6 +68,40 @@ func (r *RunStorageService) ensureBucketExists(ctx context.Context, bucket strin
...
+		// There is no worker for this notification
+		return nil
+	}
+
+	if _, err := worker.HandleTrigger(ctx, evt); err != nil {
+		return fmt.Errorf("error occcured triggering bucket notification: %w", err)
+	}
+
+	return nil
+}
```
tjholm: Should probably use GetWorkers here similar to event publishing to make sure all workers recieve all events.

tjholm (changes requested):
Just a small change to publish bucket notifications to all listeners in local dev.

pkg/run/storage.go
tjholm: These should also be sent as non-blocking requests, currently triggering an API that writes a file locally will wait for all workers to be triggered. This behaviour should be asynchronous. (See the way event publishing is handled for local dev).

pkg/run/storage.go
```diff
@@ -79,7 +117,17 @@ func (r *RunStorageService) Write(ctx context.Context, bucket string, key string
...
-	return r.StorageService.Write(ctx, bucket, key, object)
+	err = r.StorageService.Write(ctx, bucket, key, object)
+	if err != nil {
+		return err
+	}
+
+	err = r.triggerBucketNotification(ctx, bucket, key, v1.BucketNotificationType_Created)
+	if err != nil {
+		fmt.Println(err.Error())
+	}
```
tjholm: The writer shouldn't care if the responders error.

this should be something like 
```go
go r.triggerBucketNotification(context.TODO(), bucket, key, v1.BucketNotificationType_Created)
```

side log the errors in the triggerBucketNotification method instead

tjholm (changes requested):

pkg/run/storage.go
HomelessDinosaur: > Should probably use GetWorkers here similar to event publishing to make sure all workers recieve all events.

There should really only be one worker that is triggered since overlapping events are illegal

pkg/run/storage.go
```diff
@@ -64,6 +68,43 @@ func (r *RunStorageService) ensureBucketExists(ctx context.Context, bucket strin
 	return err
 }
 
+func (r *RunStorageService) triggerBucketNotifications(ctx context.Context, bucket string, key string, eventType v1.BucketNotificationType) error {
```
tjholm:
```suggestion
func (r *RunStorageService) triggerBucketNotifications(ctx context.Context, bucket string, key string, eventType v1.BucketNotificationType) {
```

Remove error return type here, as we don't return errors.

tjholm (approved):

PR #486, opened 2023-05-08
ci: bump go version to 1.19 in release flow

davemooreuws:


tjholm (changes requested):
Looks like the other workflows need updating as well

tjholm (approved):

HomelessDinosaur (approved):

PR #490, opened 2023-05-09
fix: Increase maximum volume count for seaweed for local run.

tjholm:


davemooreuws (approved):

tjholm:
just need to do some quick local testing to make sure this is all good, then will merge.

davemooreuws:
I've done 2 hours of local testing and it looks good.

PR #497, opened 2023-05-19
fixes for projects using prisma orm

jyecusch:
We may remove these fixes in future and allow developers to resolve these issues themselves through custom docker files. For now though, they have no impact on projects that aren't running prisma.

Related issues can be found here:
- https://github.com/prisma/docs/issues/4365
- https://github.com/prisma/prisma/issues/16901#issuecomment-1362940774

### DIFF
--- a/pkg/runtime/typescript.dockerfile
+++ b/pkg/runtime/typescript.dockerfile
@@ -3,27 +3,30 @@ FROM node:alpine as build
 
 ARG HANDLER
 
-WORKDIR /usr/app
-
 # Python and make are required by certain native package build processes in NPM packages.
-ENV PYTHONUNBUFFERED=1
-RUN apk add --update --no-cache python3 make g++ && ln -sf python3 /usr/bin/python
-RUN python3 -m ensurepip
-RUN pip3 install --no-cache --upgrade pip setuptools
+RUN apk add g++ make py3-pip
 
 RUN yarn global add typescript @vercel/ncc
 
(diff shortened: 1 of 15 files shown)


jyecusch:
Another fix for the bundling issue could be to remove that process from our typescript container. It's unclear how much of a performance improvement that adds at runtime.

We could also move to `tsup` with `skipNodeModulesBundle: true` to avoid bundling dependencies. That would give us a single entrypoint to run without breaking dependencies like prisma.

pkg/runtime/typescript.dockerfile
```diff
@@ -14,25 +17,32 @@ COPY . .
...
-COPY --from=build "node_modules/" "node_modules/"
+COPY . .
 
-COPY --from=build lib/ /
+RUN set -ex; \
+    yarn install --production --frozen-lockfile --cache-folder /tmp/.cache; \
+    rm -rf /tmp/.cache; \
+    # prisma fix for docker installs: https://github.com/prisma/docs/issues/4365
+    # TODO: remove when custom dockerfile support is available
+    test -d ./prisma && npx prisma generate || echo "";
```
jyecusch: It's worth noting that this line _could_ be removed from our container build and developers could add `npx prisma generate` to a postInstall hook. However, since we know prisma's popularity, it seem that would introduce unneeded friction for new users of nitric.

davemooreuws (approved):

PR #503, opened 2023-05-22
fix issues with stack commands

jyecusch:


jyecusch:
@tjholm should we move the pass phase detection to the individual providers? Since providers that don't use Pulumi probably won't need that value it feels appropriate for them to check for it themselves on start.

pkg/cmd/stack/root.go
```diff
@@ -68,6 +68,9 @@ nitric stack list
 		currPassFile := os.Getenv("PULUMI_CONFIG_PASSPHRASE_FILE")
 		if currPass == "" && currPassFile == "" {
 			p, err := preferences.GetLocalPassPhraseFile()
+			if err != nil {
+				err = fmt.Errorf("unable to determine configured pass phrase. See https://nitric.io/docs/guides/github-actions#configuring-environment-variables")
```
HomelessDinosaur:
```suggestion
				err = fmt.Errorf("unable to determine configured passphrase. See https://nitric.io/docs/guides/github-actions#configuring-environment-variables")
```

pkg/cmd/stack/root.go
```diff
@@ -218,12 +221,12 @@ var stackDeleteCmd = &cobra.Command{
...
 # To not be prompted, use -y
-nitric stack down -e aws -y`,
+nitric stack down -s aws -y`,
 	Run: func(cmd *cobra.Command, args []string) {
-		if !confirmDown {
+		if !confirmDown && !output.CI {
 			confirm := ""
 			err := survey.AskOne(&survey.Select{
-				Message: "Warning - This operation will destroy your stack, all deployed resources will be removed. Are you sure you want to proceed?",
+				Message: "Warning - This operation will destroy your stack and all resources, it cannot be undo. Continue?",
```
HomelessDinosaur:
```suggestion
				Message: "Warning - This operation will destroy your stack and all resources, it cannot be undone. Continue?",
```

pkg/preferences/userpreferences.go
```diff
@@ -53,7 +54,11 @@ type FeedbackPreferences struct {
 
 func GetLocalPassPhraseFile() (string, error) {
 	if _, err := os.Stat(utils.NitricLocalPassphrasePath()); errors.Is(err, os.ErrNotExist) {
-		err = os.WriteFile(utils.NitricLocalPassphrasePath(), []byte(defaultBackendPassPhrase), os.ModePerm)
+		// In non-CI environments we can generate the file to save a step.
+		// in CI environments this file would typically be lost, so it shouldn't auto-generate.
+		if !output.CI {
```
HomelessDinosaur: If not having the passphrase in CI will cause an error, maybe this should instead throw the error. That way the user can be directed to the docs instead of experiencing an obscure error down the line.

tjholm:
> @tjholm should we move the pass phase detection to the individual providers? Since providers that don't use Pulumi probably won't need that value it feels appropriate for them to check for it themselves on start.

Makes sense. At the moment the providers don't contain any prompting ability, so we can move it to the nitric provider section of the CLI for now until we design an API that can interact bidirectionally with the CLI

pkg/preferences/userpreferences.go
```diff
@@ -53,7 +54,11 @@ type FeedbackPreferences struct {
...
+			err = os.WriteFile(utils.NitricLocalPassphrasePath(), []byte(defaultBackendPassPhrase), os.ModePerm)
+		}
```
tjholm:
```suggestion
		}
		
```

pkg/preferences/userpreferences.go
jyecusch: Updated to separate generation from finding the file and lift the decision about whether to generate or not to the individual command that has the context about the environment it's running in.

davemooreuws (commented):
Looks good to me, just some linting errors.

davemooreuws (approved):

PR #509, opened 2023-05-30
fix: dashboard history fix

davemooreuws:


jyecusch (dismissed):

jyecusch:
What causes this error to be returned, could a more useful error message be provided so users can fix the issue preventing the history from being written?

HomelessDinosaur (dismissed):

HomelessDinosaur (approved):

jyecusch (approved):

PR #519, opened 2023-06-05
feat: Custom dockerfile support

tjholm:
Add ability to override default runtime dockerfile with own implementation. This differs from custom container support, as it still uses glob matching to produce one image per matched handler file/entrypoint.

This will allow better support for custom builds and supporting dependencies that require additional scripts to be run as part of their install processes e.g. prisma.

Example config:

```yaml
name: ts-hello
handlers:
  - match: functions/*.ts
    type: default
    docker:
      # All functions in the functions directory will be build using this dockerfile
      file: ./docker/node.dockerfile
      args: {}
preview-features:
   - dockerfile # Enable preview feature
```

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -31,7 +31,6 @@ import (
 
 	"github.com/nitrictech/cli/pkg/containerengine"
 	"github.com/nitrictech/cli/pkg/project"
-	"github.com/nitrictech/cli/pkg/runtime"
 )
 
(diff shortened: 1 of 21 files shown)


jyecusch (approved):

jyecusch:
draft docs added here: https://github.com/nitrictech/docs/pull/417

davemooreuws (approved):
LGTM just code cov check

PR #531, opened 2023-06-27
fix(python): fix occasional missing logs on AWS

jyecusch:


HomelessDinosaur (approved):

HomelessDinosaur:
We could also document that the user can add PYTHONUNBUFFERED=TRUE to their .env so it's opt-in instead of set. Can link to this issue: https://github.com/aws/amazon-sagemaker-examples/issues/319

jyecusch:
> We could also document that the user can add PYTHONUNBUFFERED=TRUE to their .env so it's opt-in instead of set. Can link to this issue: [aws/amazon-sagemaker-examples#319](https://github.com/aws/amazon-sagemaker-examples/issues/319)

True, another option would be for our SDK to call `sys.stdout.flush()` before exiting so we make sure logs are written.

That said, unless heavy logging is taking place this should not present a significant performance issue and it ensures the logs aren't lost (which is what happens currently in some cases). Maybe we should find a way to make this opt out in a future update

tjholm:
would the best of both worlds here be to set 

```dockerfile
ENV PYTHONUNBUFFERED=TRUE
```
> in the dockerfile

instead of the flag? This gives users the option of opting out in a standard way but allows us to set the default

tjholm (approved):

PR #571, opened 2023-08-28
fix: allow uppercase letters in handlers

davemooreuws:


pkg/build/build.go
```diff
@@ -102,6 +103,14 @@ func BuildBaseImages(s *project.Project) error {
 		maxConcurrency = newVal
 	}
 
+	// check functions for valid names
+	for _, fun := range s.Functions {
+		_, err := reference.Parse(fun.Name)
+		if err != nil {
+			return fmt.Errorf("function name for handler \"%s\" is invalid, must be valid ASCII containing lowercase and uppercase letters, digits, underscores, periods and hyphens", fun.Handler)
```
jyecusch:
```suggestion
			return fmt.Errorf("invalid handler name \"%s\". Names can only include alphanumeric characters, underscores, periods and hyphens", fun.Handler)
```

May as well improve the error message at the same time. Here's a suggestion to make it accurate and concise, open to alternatives.

jyecusch (changes requested):
Looks good, tested and works as expected. I made one suggestion on how to improve the error message.

pkg/build/build.go
davemooreuws: I agree with the change

jyecusch (approved):

PR #572, opened 2023-08-28
fix: Proxy local upload requests.

tjholm:


pkg/run/storage.go
```diff
@@ -159,6 +159,11 @@ func (r *RunStorageService) PreSignUrl(ctx context.Context, bucket string, key s
 		return "", err
 	}
 
+	if operation == storage.WRITE {
+		// return proxy address
+		return fmt.Sprintf("http://localhost:4000/upload/%s/%s", bucket, key), nil
```
tjholm, line 164: TODO: Need to get the gateways service address without creating a cyclic dependency (GW will depend on the storage plugin to perform the write).

PR #575, opened 2023-08-31
checks for outdated cli version

davemooreuws:
Adds check for outdated version every 24 hours on all command post runs.

### DIFF
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -1,0 +1,114 @@
+// Copyright Nitric Pty Ltd.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package update
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/pterm/pterm"
+	"github.com/spf13/cobra"
+
+	"github.com/nitrictech/cli/pkg/utils"
+)
+
+func FetchLatestVersion() string {
+	latestVersionContents, err := os.ReadFile(cachePath())
+	latestVersion := string(latestVersionContents)
+	// if file does not exist or cache is expired, fetch and save latest version
+	if err != nil || cacheExpired() {
+		owner := "nitrictech"
+		repo := "cli"
+		apiURL := fmt.Sprintf("https://api.github.com/repos/%s/%s/releases/latest", owner, repo)
+
+		response, err := http.Get(apiURL)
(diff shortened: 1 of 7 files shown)


pkg/cmd/run/root.go
```diff
@@ -70,7 +70,7 @@ var runCmd = &cobra.Command{
 			utils.CheckErr(err)
 		}
 
-		dash, err := dashboard.New(proj, envMap, true)
```
HomelessDinosaur: Is this removing the changes from the auto open browser PR?

tjholm:
One question is whether we want this post or pre-run as well. I think both are fine, although pre-run message could get lost by verbose output. I think many CLIs produce the warning pre-run as to avoid confusion with potential warning output from actual command execution though.

pkg/cmd/run/root.go
davemooreuws: yep, that should not be in there. Fixed

pkg/update/update.go
```diff
@@ -0,0 +1,132 @@
...
+func shouldSkip() (bool, error) {
+	fileInfo, err := os.Stat(cachePath())
+	if err != nil {
+		// no file, create and return false
+		err = updateFile()
+
+		return false, err
+	}
+
+	if time.Now().After(fileInfo.ModTime().Add(24 * time.Hour)) {
```
HomelessDinosaur: I think it's fine to show the version warning every time. Maybe just update the cache every day?

pkg/update/update.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	err = json.Unmarshal(body, &releaseInfo)
+	if err != nil {
+		cobra.CheckErr(err)
+	}
+
+	LatestVersion = strings.TrimPrefix(releaseInfo.TagName, "v")
+
+	shouldShowWarning = CurrentVersion < LatestVersion
```
tjholm: function `ShowWarning` should not rely on global state. Calculate `shouldShowWarning` where its being used.

pkg/update/update.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/nitrictech/cli/pkg/utils"
+)
+
+var (
+	shouldShowWarning bool
+	CurrentVersion    string
+	LatestVersion     string
+)
```
tjholm: Lets avoid using global state where we can.

The workflow for this could probably be:

1. FetchLatestVersion() - This will fetch and cache the latest version of nitric (once per day to avoid github API throttling). This result will be stored like is already happening. This method can probably just return the latest version from cache as well (and only refresh when stale).

2. `isOutdated()` - Can probably be private for now, will calculate if the version is out of date based on the cache from `FetchLatestVersion()`.

3. `PrintOutdatedWarning()` - Prints the warning if `isOutdated() === true`.

If we want to use global variables as a cache we should make sure they're only every written/read from a single method.

pkg/update/update.go
```diff
@@ -0,0 +1,132 @@
...
+		cobra.CheckErr(err)
+	}
+
+	LatestVersion = strings.TrimPrefix(releaseInfo.TagName, "v")
+
+	shouldShowWarning = CurrentVersion < LatestVersion
+}
+
+func ShowWarning() {
+	if shouldShowWarning {
```
tjholm: I wouldn't worry about this too much, if this is possible we've either included `ShowWarning()` in the wrong place or command code is too `DRY`. If this is to stop a second printing for the exit for our blocking commands like `start` and `run`. Let's exclude these and work on the banner in the dashboard to patch this up :).

tjholm (approved):

PR #576, opened 2023-08-31
feat: auto open local dashboard on nitric start

davemooreuws:
On start command, opens browser on localhost:port. Added --no-browser flag to start command to avoid this happening.

Tested on Windows. Please test on linux and macos


tjholm:
Might want to factor in the CI mode as well to avoid opening for users running `nitric start` as part of their CI/CD pipelines for e2e testing.

HomelessDinosaur (approved):

PR #582, opened 2023-09-12
fix: no need to validate handlers on nitric start

davemooreuws:


pkg/cmd/start/root.go
```diff
@@ -57,17 +57,22 @@ var startCmd = &cobra.Command{
...
-		utils.CheckErr(err)
+		history := &history.History{
+			ProjectDir: config.Dir,
+		}
+
+		proj := &project.Project{
+			Name:    config.Name,
+			Dir:     config.Dir,
+			History: history,
+		}
```
tjholm, line 68: Does History need to be apart of project? Looking at the references all access to History are as a public field on Project.

pkg/cmd/start/root.go
```diff
@@ -57,17 +57,22 @@ var startCmd = &cobra.Command{
...
 		dash, err := dashboard.New(proj, map[string]string{}, noBrowser || output.CI)
 		utils.CheckErr(err)
 
 		ls := run.NewLocalServices(&project.Project{
-			Name: "local",
-			History: &history.History{
-				ProjectDir: proj.Dir,
-			},
+			Name:    "local",
+			History: history,
```
tjholm: This constructs a new reference to project, can the existing reference be used? Or is this a result of history needing to be apart of project? (If they were seperated would we still need Project here?)

davemooreuws: Not sure why the name is local here instead of the project name, any idea? That threw me off

pkg/cmd/start/root.go
davemooreuws: yes otherwise we get a nil pointer error elsewhere. Could probably be fixed by making more methods detached, this was less churn in the PR changes before our refactor tasks

pkg/cmd/start/root.go
tjholm: This looks like filler to accomodate for History without providing a History object directly (I haven't traced it but on its face it appears that way.)

pkg/cmd/start/root.go
davemooreuws: Discussed, this will be considered in the refactor

tjholm (approved):

PR #591, opened 2023-10-11
feat: update 'stack new' cmd UX

### LINKED ISSUE #506: cli: improve new stack command DX
Update DX and prompts in the `nitric stack new` command. Let's update it to match the new DX of the new project command, with clear prompts for each step and inline input validation to minimize errors causing a restart of the process.

davemooreuws:
Closes [506](https://github.com/nitrictech/nitric/issues/506)


README.md
```diff
@@ -37,7 +37,7 @@ Common commands in the CLI that you’ll be using:
 
 - nitric down : Undeploy a previously deployed stack, deleting resources
 - nitric run : Run your project locally for development and testing
-- nitric stack new : Create a new Nitric stack
+- nitric stack new [stackName] [cloudName] [region] : Create a new Nitric stack
```
tjholm: Not sure if we should include region at the top level as its varies too much by cloud.

HomelessDinosaur:
Getting this issue, probably ignore if the provider exists at this stage as it gets downloaded during the deployment stage. There shouldn't be a scenario where the default provider we are choosing for the stack file is invalid.
Screen Shot 2023-11-14 at 12 54 04 pm

HomelessDinosaur (approved):

tjholm (approved):

PR #598, opened 2023-11-16
fix: avoid history contention

davemooreuws:
Avoids history file write contention. Also includes a few other small fixes.

### DIFF
--- a/pkg/history/history.go
+++ b/pkg/history/history.go
@@ -22,8 +22,11 @@ import (
 	"io/fs"
 	"os"
 
+	"github.com/nitrictech/cli/pkg/eventbus"
 	"github.com/nitrictech/cli/pkg/utils"
 )
+
+const AddRecordTopic = "history:addrecord"
 
 type HistoryRecords struct {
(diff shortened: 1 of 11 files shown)


pkg/history/history.go
```diff
@@ -88,41 +88,77 @@ type ResponseHistory struct {
...
-
-	err = os.WriteFile(historyFile, data, fs.ModePerm)
-	if err != nil {
-		return NewHistoryError(recordType, historyFile)
-	}
+	// Enqueue the write operation
+	h.writeQueue <- writeOp
 
 	return nil
 }
```
tjholm: Does the channel need to accept callbacks which have their own slices of callbacks? Can the history writer just accept history entries as structs? And the function that's queuing it can treat it as successful once queued?

tjholm (changes requested):

pkg/history/history.go
davemooreuws: See new updates

tjholm (approved):

HomelessDinosaur (approved):

PR #604, opened 2023-11-23
feat: add command for previewing deployments

HomelessDinosaur:
Relies on: https://github.com/nitrictech/nitric/pull/543


cmd/stack.go
```diff
@@ -176,6 +177,44 @@ nitric stack down -s aws -y`,
 	Args: cobra.ExactArgs(0),
 }
 
+var stackPreviewCommand = &cobra.Command{
+	Use:     "preview [-s stack]",
+	Short:   "Preview the deployment of a stack",
+	Long:    `Preview the deployment of a stack`,
+	Example: `nitric stack down -s aws`,
```
davemooreuws:
```suggestion
	Example: `nitric stack preview -s aws`,
```

cmd/stack.go
```diff
@@ -176,6 +177,44 @@ nitric stack down -s aws -y`,
...
+	Run: func(cmd *cobra.Command, args []string) {
+		s, err := stack.ConfigFromOptions()
+
+		if err != nil && strings.Contains(err.Error(), "No nitric stacks found") {
```
davemooreuws: This block doesn't work for me. Yes sends the error "No nitric stacks found".
Same thing happens to the up command. Could be an unrelated issue to this PR

cmd/stack.go
```diff
@@ -185,6 +224,11 @@ func init() {
 	stackUpdateCmd.Flags().BoolVarP(&forceStack, "force", "f", false, "force override previous deployment")
 	utils.CheckErr(stack.AddOptions(stackUpdateCmd, false))
 
+	stackCmd.AddCommand(command.AddDependencyCheck(stackPreviewCommand, command.Pulumi, command.Docker))
+	stackPreviewCommand.Flags().StringVarP(&envFile, "env-file", "e", "", "--env-file config/.my-env")
+	stackPreviewCommand.Flags().BoolVarP(&forceStack, "force", "f", false, "force override previous deployment")
```
davemooreuws: Does preview need a force?

davemooreuws (changes requested):

cmd/stack.go
HomelessDinosaur: If its in a conflicting state then it needs a force

cmd/stack.go
HomelessDinosaur: This was removed from `up` as well when it was refactored, there are a few lines commented out. Will update now.

davemooreuws (dismissed):

HomelessDinosaur:
Marking as draft until the Nitric provider PR gets merged.

PR #609, opened 2023-12-28
fix: deduplicate resources when creating policies

HomelessDinosaur:


tjholm:
This change looks looks fine, just need to make sure that it deduplicates on the same level as the original implementation. The reason is that there is an policy application limit for AWS Lambda accounts that this was implemented to address, but it appears that this change may undo that.

The intention here was to ensure that all resource types that shared a common set of actions were grouped into a single policy, so in the case of something like read-only secrets we could merge them to save on IAM policy length.

e.g.

functionA:read:secret1 + functionA:read:secret2 + functionA:read:secret3 => functionA:read:secret1,secret2,secret3

PR #654, opened 2024-02-12
notify of errors during container run

tjholm:
Currently errors on the SDK side will be silent, this raises those errors during collections/runs.

Output will look something like:
image

### DIFF
--- a/pkgplus/project/service.go
+++ b/pkgplus/project/service.go
@@ -1,6 +1,7 @@
 package project
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


davemooreuws (approved):

pkgplus/project/service.go
```diff
@@ -398,12 +400,38 @@ func (s *Service) RunContainer(stop <-chan bool, updates chan<- ServiceRunUpdate
...
+
+				err = fmt.Errorf("service %s exited with non 0 status\n %s", s.Name, logs.String())
+
+				updates <- ServiceRunUpdate{
+					ServiceName: s.Name,
+					// TODO: Extract the error logs for the container here...
+					Err:    err,
+					Status: ServiceRunStatus_Error,
+				}
+				return err
```
tjholm, line 426: If we want to handle this in a tea program instead, just need to make this return nil and have the tea program handle the updates channel.

jyecusch (approved):

PR #662, opened 2024-02-15
don't panic on docker daemon not running

davemooreuws:


cmd/run.go
```diff
@@ -65,6 +66,10 @@ var runCmd = &cobra.Command{
 		go func() {
 			err := proj.RunServices(localCloud, stopChan, updatesChan)
 			if err != nil {
+				if strings.Contains(err.Error(), "Docker daemon is not running") {
+					tui.CheckErr(err)
+				}
```
tjholm: we might be able to check this at the start instead. Or use the command dependencies to ensure everything is running?

jyecusch (approved):

tjholm (approved):

PR #669, opened 2024-02-16
add pulumi passphrase file generation

jyecusch:


cmd/stack.go
```diff
@@ -177,6 +178,10 @@ var stackUpdateCmd = &cobra.Command{
 		stackConfig, err := stack.ConfigFromName[map[string]any](fs, stackSelection)
 		tui.CheckErr(err)
 
+		if !isNonInteractive() {
+			_ = pulumi.EnsurePulumiPassphrase(fs)
+		}
```
tjholm, line 183: Did we still want to check for a nitric provider in this case?

tjholm (approved):

PR #678, opened 2024-02-20
v1 websocket edge updates

davemooreuws:


jyecusch (approved):

pkg/dashboard/frontend/src/lib/utils/generate-architecture-data.ts
```diff
@@ -255,25 +255,6 @@ export function generateArchitectureData(data: WebSocketResponse): {
...
-            type: MarkerType.ArrowClosed,
-          },
-          markerStart: {
-            type: MarkerType.ArrowClosed,
-            orient: 'auto-start-reverse',
-          },
-          label: eventType,
-        }
-      }),
-    )
```
tjholm: This just outright deletes the Triggering edge. Do we want to get rid of these and only display the manage policy for the websockets? Because policies are rendered last maybe we could check if the service is already a target of the resource it has permission to and merge there?

tjholm (dismissed):

PR #696, opened 2024-02-29
fix: remove pterm with tui

davemooreuws:
On windows git bash and potentially other environments, pterm causes colors to break in the terminal output.

We shouldn't be mixing pterm with lipgloss/bubbletea. This PR replaces all pterm calls with a tui view.

Looks like this:
image


cmd/root.go
```diff
@@ -85,7 +80,7 @@ var rootCmd = &cobra.Command{
 func Execute() {
 	defer func() {
 		if err := recover(); err != nil {
-			pterm.Error.Printfln(
+			_, _ = tui.Error.Printfln(
```
HomelessDinosaur: Does this need to be assigned?
```suggestion
			tui.Error.Printfln(
```

HomelessDinosaur (dismissed):

cmd/root.go
davemooreuws: was getting a weird lint error if I didnt 👯

tjholm: this could be addressed by not returning the result in the prefix printer methods. I'm pretty sure the `fmt` package has exceptions for error/assignment handling.

davemooreuws: Pushed changes

tjholm (approved):

HomelessDinosaur (approved):

PR #698, opened 2024-03-01
fix: filter out conflicting os vars

davemooreuws:
apis are not being deployed correctly from windows due to conflicting os env vars being sent into the container. This could be an issue anywhere.


HomelessDinosaur:
If it's only a windows issue, should you only filter if windows is detected? Also maybe debug log which environment variables are being filtered out, might cause issues if a dev expects it to exist.

davemooreuws:
> If it's only a windows issue, should you only filter if windows is detected? Also maybe debug log which environment variables are being filtered out, might cause issues if a dev expects it to exist.

Could potentially happen on other platforms. Windows it just happens OOTB.

HomelessDinosaur (approved):

tjholm (approved):

PR #703, opened 2024-03-17
fix: ensure docker runtimes support dockerignore, empty build args and ignore other entrypoint files

davemooreuws:
- Stops the CLI from breaking when no args are defined
- Allows .dockerignore to be used for default runtimes
- Allows a custom runtime to define a dockerignore by "[dockerfile]".dockerignore
- Adds other entrypoint files to dockerignore

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -309,6 +309,10 @@ func fromProjectConfiguration(projectConfig *ProjectConfiguration, fs afero.Fs) 
 
 			var buildContext *runtime.RuntimeBuildContext
 
+			otherEntryPointFiles := lo.Filter(files, func(file string, index int) bool {
+				return file != f
+			})
+
 			if serviceSpec.Runtime != "" {
 				// We have a custom runtime
 				customRuntime, ok := projectConfig.Runtimes[serviceSpec.Runtime]
@@ -320,8 +324,7 @@ func fromProjectConfiguration(projectConfig *ProjectConfiguration, fs afero.Fs) 
 					relativeServiceEntrypointPath,
 					customRuntime.Dockerfile,
 					customRuntime.Args,
-					// TODO: Get other entrypoint files as ignores
-					[]string{},
+					otherEntryPointFiles,
 					fs,
 				)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/project/runtime/runtime.go
```diff
@@ -55,7 +55,12 @@ func customBuildContext(entrypointFilePath string, dockerfilePath string, buildA
 		return nil, err
 	}
 
-	// Get the ignore file contents
+	// TODO Get the ignore file contents
```
HomelessDinosaur: How difficult is this to add? I think we were avoiding TODO statements after v1

davemooreuws: I've now added the dockerignore logic into this PR

pkg/project/project.go
```diff
@@ -320,8 +320,6 @@ func fromProjectConfiguration(projectConfig *ProjectConfiguration, fs afero.Fs)
 					relativeServiceEntrypointPath,
 					customRuntime.Dockerfile,
 					customRuntime.Args,
-					// TODO: Get other entrypoint files as ignores
-					[]string{},
```
tjholm, line 324: This TODO is actually important, we need the ability to remove additional files outside of the scope of the .dockerignore file so we can do things like removing entrypoints from the scope of other entrypoint files/services.

tjholm (dismissed):

pkg/project/project.go
davemooreuws: This has now been added to the PR

pkg/project/runtime/runtime.go
```diff
@@ -47,6 +48,31 @@ const (
...
+
+		// Split the content into lines
+		lines := lo.Filter[string](strings.Split(string(content), "\n"), func(line string, index int) bool {
+			return strings.TrimSpace(line) != ""
+		})
+
+		return lines, nil
+	}
+
+	return nil, nil
```
tjholm: return default value over nil, or an error if not exists and handle outside? double nil returns can be confusing and error prone (requires reading the code to understand the case, when you encounter a nil pointer exception from misuse).

davemooreuws: changed to use default

HomelessDinosaur (approved):

tjholm (approved):

PR #723, opened 2024-04-17
feat: add check for leading slash

HomelessDinosaur:


davemooreuws (approved):
Looks good to me.

> Note: for node we can add the type catching in TypeScript.

tjholm (approved):

PR #735, opened 2024-06-12
feat: ability to lint and validate the nitric.yaml from json schema

davemooreuws:
Note: Used a simple lock file for version check instead of parsing the JSON each time.

Should test against a published version of vscode ext before merging.


tjholm:
closing as stale

PR #739, opened 2024-06-21
fix: remove the default golang runtime

tjholm:
Golang runtime to be shipped with templates.


pkg/project/runtime/runtime.go
```diff
@@ -220,8 +204,11 @@ func NewBuildContext(entrypointFilePath string, dockerfilePath string, buildArgs
 		return customBuildContext(entrypointFilePath, dockerfilePath, buildArgs, additionalIgnores, fs)
 	}
 
-	ext := filepath.Ext(entrypointFilePath)
+	if fi, err := fs.Stat(entrypointFilePath); err == nil && fi.IsDir() {
+		return nil, fmt.Errorf("nitric does not support directories by default, use a custom runtime with a Dockerfile")
```
davemooreuws: Should we provide a link?

tjholm: doc link in

jyecusch (approved):

HomelessDinosaur (approved):

PR #745, opened 2024-07-11
feat: add ability to set build context for custom runtimes

### LINKED ISSUE #744: Improve support for monorepos with docker context config
### Suggestion
At the moment the docker context is wherever the nitric project is located. A solution to this would be to add a new key to the custom runtimes config called `context`

### Value
Ability to include all the files for a monorepo, to allow it to be installed and built and share packages.

### Alternatives
Moving projects to the root. Which isn't ideal.

## Example for turbo repo

```yaml
name: guestbook-api
services:
  - match: services/*.ts
    runtime: turbo
    type: ""
    start: npm run dev:services $SERVICE_PATH
runtimes:
  turbo:
    # All services that specify the 'turbo' runtime will be built using this dockerfile
    dockerfile: ./turbo.dockerfile
    context: ../../
    args:
      TURBO_SCOPE: "guestbook-api"
```

davemooreuws:
Closes #744

### DIFF
--- a/pkg/project/config.go
+++ b/pkg/project/config.go
@@ -30,6 +30,8 @@ import (
 type RuntimeConfiguration struct {
 	// Template dockerfile to use as the base for the custom runtime
 	Dockerfile string
+	// Directory path for the Docker build context for the custom runtime
+	Context string
 	// Additional args to pass to the custom runtime
 	Args map[string]string
 }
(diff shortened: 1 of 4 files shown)


tjholm (dismissed):

cmd/run.go
```diff
@@ -97,6 +97,7 @@ var runCmd = &cobra.Command{
 		go func() {
 			err := proj.RunServices(localCloud, stopChan, updatesChan)
 			if err != nil {
+				localCloud.Stop()
 				panic(err)
 			}
 		}()
```
tjholm: Can address this in a separate PR, but we might want to actually capture and return the errors here instead of panicking.

tjholm (approved):

HomelessDinosaur (approved):

PR #746, opened 2024-07-11
feat: allow HTTPS for local APIs and Http Proxies

jyecusch:
Enabled using `nitric start --https-preview`

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -69,7 +69,7 @@ var runCmd = &cobra.Command{
 		}
 
 		var tlsCredentials *gateway.TLSCredentials
-		if httpsExperimental {
+		if enableHttps {
 			createTlsCredentialsIfNotPresent(fs, proj.Directory)
 			tlsCredentials = &gateway.TLSCredentials{
 				CertFile: paths.NitricTlsCertFile(proj.Directory),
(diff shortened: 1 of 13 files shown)


cmd/run.go
```diff
@@ -75,7 +75,7 @@ var runCmd = &cobra.Command{
 		var localCloud *cloud.LocalCloud
 		go func() {
 			// Start the local cloud service analogues
-			localCloud, err = cloud.New(proj.Name)
+			localCloud, err = cloud.New(proj.Name, cloud.LocalCloudOptions{})
```
tjholm: Are we also planning on adding SSL support for `run`

algoflows: How can I help test this baddy, would be awesome to have this merged as I could use this feature for my project at work, any ETA on when this might be live on prod?

tjholm: @algoflows just pushed changes to also implement this for `nitric run`, so should be released shortly.

tjholm (dismissed):

tjholm (dismissed):

pkg/cloud/gateway/gateway.go
```diff
@@ -124,7 +145,14 @@ func (s *LocalGatewayService) GetHttpWorkerAddresses() map[string]string {
 
 	if len(s.httpServers) > 0 && len(s.httpWorkers) == len(s.httpServers) {
 		for idx, host := range s.httpWorkers {
-			addresses[host] = strings.Replace(s.httpServers[idx].lis.Addr().String(), "[::]", "localhost", 1)
+			protocol := "http"
+			if s.apiServers[idx].tlsCredentials != nil {
```
davemooreuws:
```suggestion
			if s.httpServers[idx].tlsCredentials != nil {
```

davemooreuws (changes requested):

davemooreuws:
Looks good just one change.

davemooreuws (approved):

tjholm (approved):

PR #753, opened 2024-07-24
fix: require docker for nitric down

davemooreuws:
Fixes an issue where a pulumi refresh requires docker

```
rpc error: code = Unknown desc = failed to refresh stack: exit status 0xffffffff

...

Diagnostics:
  pulumi:pulumi:Stack (demo-demo-pulumi):
    error: update failed

  docker:index:Image (demo_services-photos-image):
    error: Docker native provider returned an unexpected error from Configure: failed to connect to any docker daemon
```

### DIFF
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -627,7 +627,7 @@ func init() {
 	tui.CheckErr(AddOptions(stackUpdateCmd, false))
 
 	// Delete Stack (Down)
-	stackCmd.AddCommand(tui.AddDependencyCheck(stackDeleteCmd, tui.Pulumi))
+	stackCmd.AddCommand(tui.AddDependencyCheck(stackDeleteCmd, tui.Pulumi, tui.Docker))
 	stackDeleteCmd.Flags().BoolVarP(&confirmDown, "yes", "y", false, "confirm the destruction of the stack")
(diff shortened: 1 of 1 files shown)


jyecusch (approved):

cmd/stack.go
```diff
@@ -627,7 +627,7 @@ func init() {
 	tui.CheckErr(AddOptions(stackUpdateCmd, false))
 
 	// Delete Stack (Down)
-	stackCmd.AddCommand(tui.AddDependencyCheck(stackDeleteCmd, tui.Pulumi))
+	stackCmd.AddCommand(tui.AddDependencyCheck(stackDeleteCmd, tui.Pulumi, tui.Docker))
```
tjholm, line 630: Now that I think about this with the release of our terraform providers we'll want to shift more of this to our providers, possibly implementing somekind of preflight check for dependencies, so our terraform users don't require pulumi to be installed simply to execute the commands.

tjholm (approved):

cmd/stack.go
davemooreuws: Created an issue here: https://github.com/nitrictech/nitric/issues/649

PR #770, opened 2024-08-13
fix: ensure docker in docker can run properly via new env variable

### LINKED ISSUE #769: GitLab pipelines issue running nitric up on shared runners
### Issue

On GitLab pipelines, with shared runners there are issues connecting to the containers.

### Steps

With latest docker (27~) there are connection errors.

```yaml
deploy:
  image: docker:latest
  services:
    - docker:dind
  rules:
    - if: $CI_PIPELINE_SOURCE == 'merge_request_event'
    - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
  variables:
    PULUMI_CONFIG_PASSPHRASE: $PULUMI_ACCESS_TOKEN
    PULUMI_ACCESS_TOKEN: $PULUMI_ACCESS_TOKEN
    AWS_ACCESS_KEY_ID: $AWS_ACCESS_KEY_ID
    AWS_SECRET_ACCESS_KEY: $AWS_SECRET_ACCESS_KEY

  before_script:
    - apk update && apk add --no-cache curl bash
    - curl -L https://nitric.io/install?version=latest | bash
    - export PATH=$PATH:$HOME/.nitric/bin
    - curl -fsSL https://get.pulumi.com/ | bash
    - export PATH=$PATH:$HOME/.pulumi/bin
  script:
    - nitric up --ci
```

With docker in docker recommended version docker (24~) there are resources that are not deployed. (i.e. API gateway)

```yaml
deploy:
  image: docker:24.0.5
  services:
    - docker:24.0.5-dind
  rules:
    - if: $CI_PIPELINE_SOURCE == 'merge_request_event'
    - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
  variables:
    PULUMI_CONFIG_PASSPHRASE: $PULUMI_ACCESS_TOKEN
    PULUMI_ACCESS_TOKEN: $PULUMI_ACCESS_TOKEN
    AWS_ACCESS_KEY_ID: $AWS_ACCESS_KEY_ID
    AWS_SECRET_ACCESS_KEY: $AWS_SECRET_ACCESS_KEY

  before_script:
    - apk update && apk add --no-cache curl bash
    - curl -L https://nitric.io/install?version=latest | bash
    - export PATH=$PATH:$HOME/.nitric/bin
    - curl -fsSL https://get.pulumi.com/ | bash
    - export PATH=$PATH:$HOME/.pulumi/bin
  script:
    - nitric up --ci
```

I have also tried adding these variables, which do not fix any of the above:

```
DOCKER_HOST: tcp://docker:2375
DOCKER_DRIVER: overlay2
DOCKER_TLS_CERTDIR: ""
```

### Expected

It should work.

### Environment and setup information

- Nitric CLI Version: v1.50.2
- Operating System:
- Nitric dependencies and their versions, such as SDKs or Middleware: Python SDK
- Cloud providers you are deploying to and their version: nitric/aws@1.11.6

### Other info
Errors with latest docker:
image

False success with docker 24 (missing API gateway output/deployment):
image

davemooreuws:
Fixes #769 

Adds new env variable called `NITRIC_DOCKER_HOST` , this can be set to an IP like `172.17.0.4` on runners like GitLab and other Docker in Docker environments.

I have tested on:
- GitLab runners with docker 27


pkg/docker/linuxhost.go
```diff
@@ -0,0 +1,64 @@
...
+		if iface.Name == "docker0" {
+			return getNonLoopback(iface)
+		}
+	}
+
+	for _, iface := range interfaces {
+		if iface.Name == "eth0" {
+			return getNonLoopback(iface)
+		}
+	}
```
tjholm:
```suggestion
        ifaceNames := []string{ "docker0", "eth0" }
        for _, name := range ifaceNames {
            for _, iface := range interfaces {
                if iface.Name == name {
                    return getNonLoopback(iface)
                }
            }
        } 
```

tjholm (dismissed):

tjholm:
It looks like gitlab exposes a `docker` hostname locally.

`DOCKER_HOST` as you've mentioned appears to be a convention for configuring this in CI/CD.

When you tested setting `DOCKER_HOST` did you use the env variable in the CLI to set `host.docker.internal` to the same value?

Rather than searching the `eth0` which may not apply in all cases we could look at allowing it to be configured? e.g. use `DOCKER_HOST` or something else like `NITRIC_HOST`?

Main concern is that a search for `eth0` will not cover all cases or may change/break in ways we may not be able to control or predict.

davemooreuws:
> It looks like gitlab exposes a `docker` hostname locally.
> 
> 
> 
> `DOCKER_HOST` as you've mentioned appears to be a convention for configuring this in CI/CD.
> 
> 
> 
> When you tested setting `DOCKER_HOST` did you use the env variable in the CLI to set `host.docker.internal` to the same value?
> 
> 
> 
> Rather than searching the `eth0` which may not apply in all cases we could look at allowing it to be configured? e.g. use `DOCKER_HOST` or something else like `NITRIC_HOST`?
> 
> 
> 
> Main concern is that a search for `eth0` will not cover all cases or may change/break in ways we may not be able to control or predict.

Agreed, we want to avoid it being brittle. Will need to double check setting the same value as DOCKER_HOST. I think a nitric based env var will work more broadly (have a feeling the azure and GCP might need configuring as well), but let's try it tomorrow. We can keep the WSL fix in.

jyecusch (approved):

tjholm (approved):

PR #771, opened 2024-08-15
fix: issues executing multiple sql queries in dashboard

davemooreuws:


HomelessDinosaur (dismissed):

pkg/cloud/sql/sql.go
```diff
@@ -193,38 +193,30 @@ func (l *LocalSqlServer) Query(ctx context.Context, connectionString string, que
...
-		if command == "" {
-			continue
-		}
+	query = strings.TrimSpace(query)
 
-		rows, err := tx.Query(ctx, command)
-		if err != nil {
-			_ = tx.Rollback(ctx)
+	// Execute the query with simple protocol
+	rows, err := tx.Query(ctx, query, pgx.QueryExecModeSimpleProtocol)
```
HomelessDinosaur: For others reviewing:

```
// QueryExecModeSimpleProtocol should have the user application visible behavior as QueryExecModeExec with minor
// exceptions such as behavior when multiple result returning queries are erroneously sent in a single string.
//
// QueryExecModeSimpleProtocol uses client side parameter interpolation. All values are quoted and escaped. Prefer
// QueryExecModeExec over QueryExecModeSimpleProtocol whenever possible. In general QueryExecModeSimpleProtocol
// should only be used if connecting to a proxy server, connection pool server, or non-PostgreSQL server that does
// not support the extended protocol.
```

HomelessDinosaur (approved):

raksiv (approved):

PR #780, opened 2024-08-30
feat: local sql migrations

### LINKED ISSUE #660: support running database migrations on nitric start/run
### Suggestion

Support running database migrations locally using the same process that's used in deployed environments.

### Value

Validates that the migrations will work using the deployed process (e.g. all migration filenames are valid, they're all discovered, they run in the intended order, etc.)

### Alternatives

Run migrations using other tools, e.g. an ORM cli (if used) or execute the SQL files using another client. However, while this may validate the migrations work with the database, it doesn't ensure they work with the process nitric uses in the cloud.

For example, if the file prefixes are wrong, the SQL files will still work with other tools but fail when deployed.

davemooreuws:
Ability to build and run local migrations on nitric start and run via the "Run migrations" button. 

Closes https://github.com/nitrictech/nitric/issues/660


pkg/cloud/sql/sql.go
```diff
@@ -236,16 +282,135 @@ func (l *LocalSqlServer) Query(ctx context.Context, connectionString string, que
...
+			l.Publish(l.State)
+		}
+	}
+
+	if len(databasesToMigrate) > 0 {
+		err := l.BuildAndRunMigrations(databasesToMigrate)
+		if err != nil {
+			exit.GetExitService().Exit(fmt.Errorf("failed to build and run migrations: %w", err))
+		}
+	}
```
tjholm: Including automated migrations in an update handle for local resource state might be a touch aggressive? If I'm reading this right, this will trigger migrations for existing and new databases on any resource state change (e.g. Adding a Topic).

For now I think providing a convenient way to manually trigger migrations should suffice, as you've already included in the dashboard.

tjholm (dismissed):

pkg/cloud/sql/sql.go
davemooreuws: Updated

tjholm:
@davemooreuws have made changes to move service/migrations back to the project package for now. I've also changed to exit service to a system logging service that logs in the CLI along with running nitric services as the "nitric" service in order to side log runtime errors rather than exiting the CLI for now.

Can you re-test to make sure everything is working as you expect.

davemooreuws:
> @davemooreuws have made changes to move service/migrations back to the project package for now. I've also changed to exit service to a system logging service that logs in the CLI along with running nitric services as the "nitric" service in order to side log runtime errors rather than exiting the CLI for now.
> 
> Can you re-test to make sure everything is working as you expect.

Tested, works great!

tjholm (approved):

HomelessDinosaur (approved):

PR #781, opened 2024-09-02
chore: create automatic PRs for new nitric provider versions

davemooreuws:
on schedule or manual dispatch


davemooreuws:
Not required once this is released: https://github.com/nitrictech/cli/pull/814

PR #782, opened 2024-09-03
local dashboard support for jobs

davemooreuws:
Will add tests once node-sdk is released.


pkg/cloud/batch/batch.go
```diff
@@ -19,46 +19,178 @@ package batch
...
+	}
+
+	_, err = l.HandleJobRequest(&batchpb.ServerMessage{
+		Content: &batchpb.ServerMessage_JobRequest{
+			JobRequest: &batchpb.JobRequest{
+				JobName: req.GetJobName(),
+				Data:    req.Data,
 			},
+		},
+	})
```
tjholm: Job runs should be asynchronous, this change makes them synchronous. If you have a job that runs for 30mins this will block the return of the calling endpoint.

tjholm (dismissed):

PR #789, opened 2024-10-01
fix: check for valid openid configuration url

HomelessDinosaur:


pkg/collector/spec.go
```diff
@@ -444,6 +444,10 @@ func buildApiRequirements(allServiceRequirements []*ServiceRequirements, project
 						projectErrors.Add(fmt.Errorf("service %s attempted to register an OIDC security scheme with an empty issuer", serviceRequirements.serviceName))
 					}
 
+					if !strings.Contains(issuerUrl, "/.well-known/openid-configuration") {
+						projectErrors.Add(fmt.Errorf("service %s attempted to register an OIDC security policy with an invalid OIDC URL", serviceRequirements.serviceName))
+					}
```
tjholm: can we confirm that all providers conform to this as a standard? I think I remember AWS cognito being a bit different.

It might actually be worth trying to actually fetch the document and seeing if its a valid OIDC public doc.

tjholm (dismissed):

davemooreuws (changes requested):
fix lint issues

pkg/collector/openid.go
```diff
@@ -0,0 +1,93 @@
...
+
+type OpenIdConfig struct {
+	Issuer        string `json:"issuer"`
+	JwksUri       string `json:"jwks_uri"`
+	TokenEndpoint string `json:"token_endpoint"`
+	AuthEndpoint  string `json:"authorization_endpoint"`
+}
+
+func validateOpenIdConnectConfig(rawUrl string) error {
+	// append well-known configuration to issuer
```
jyecusch: This comment seems out of context since nothing appears to be appended here (is it missing?)

pkg/collector/openid.go
```diff
@@ -0,0 +1,93 @@
...
+	openIdConnectUrl, err := url.Parse(rawUrl)
+	if err != nil {
+		return err
+	}
```
jyecusch: Maybe just accept a valid URL as the input to this function e.g.

```go
validateOpenIdConnectConfig(issuerUrl *url.URL) error {
```

I think the name being more specific than rawUrl could also help the caller know what's expected

pkg/collector/openid.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return fmt.Errorf("OIDC provider failed to respond with config within %s", timeout.String())
+		}
+
+		return err
+	}
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("received non 200 status retrieving openid-configuration: %d", resp.StatusCode)
```
jyecusch: Return the body too since it might help with debugging (e.g. maybe a proxy is in the way or something else the user can fix)

pkg/collector/openid.go
```diff
@@ -0,0 +1,93 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	oidConf := &OpenIdConfig{}
+
+	if err := json.Unmarshal(body, oidConf); err != nil {
+		return errors.WithMessage(err, "error unmarshalling open id config")
+	}
```
jyecusch:
```suggestion
	var oidConf OpenIdConfig
	
	if err := json.Unmarshal(body, &oidConf); err != nil {
		return errors.WithMessage(err, "error unmarshalling open id config")
	}
```

jyecusch (dismissed):

davemooreuws (dismissed):

pkg/collector/openid.go
```diff
@@ -0,0 +1,87 @@
...
+	// get the configuration document
+	resp, err := client.Get(openIdConnectUrl.String())
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return fmt.Errorf("OIDC provider failed to respond with config within %s", timeout.String())
+		}
+
+		return err
+	}
+
```
raksiv, line 54:
```suggestion
defer resp.Body.Close()

```

pkg/collector/openid.go
```diff
@@ -0,0 +1,87 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("received %d status retrieving openid-configuration: %s", resp.StatusCode, body)
+	}
+
+	var oidcConfig *OpenIdConfig
```
raksiv: Got some nil dereferencing warnings, can avoid by using a value instead of a pointer.

```go
	var oidcConfig OpenIdConfig

	if err := json.Unmarshal(body, &oidcConfig); err != nil {
		return errors.WithMessage(err, "error unmarshalling open id config")
	}
```

raksiv (dismissed):
Looking good, some small suggestions.

raksiv (approved):

davemooreuws (approved):

PR #790, opened 2024-10-04
fix: schedules and batch docs links

jyecusch:


davemooreuws (approved):

pkg/dashboard/frontend/src/components/events/EventsExplorer.tsx
```diff
@@ -44,6 +44,12 @@ interface Props {
   workerType: 'schedules' | 'topics' | 'jobs'
 }
 
+const docsPathMap = {
+  schedules: 'schedules',
+  topics: 'topics',
```
davemooreuws:
```suggestion
  topics: 'messaging',
```

davemooreuws: topics isn't a valid link. 

Could add a temporary redirect to docs first and then do this? Then in the new docs we will add a topics page.
Or leave it as messaging for now.

davemooreuws (changes requested):

davemooreuws (approved):

HomelessDinosaur (approved):

PR #799, opened 2024-10-21
fix: Add resource name validation to local run and collection

tjholm:


davemooreuws:
@tjholm The errors show up nicely and it works well. However there is one scenario that forces the user to restart the start command.

Steps to reproduce (in the TS example):
1. Change API to an invalid name
2. Change to the suggested name
3. As it refreshes you can see the API has not been registered correctly

tjholm:
> @tjholm The errors show up nicely and it works well. However there is one scenario that forces the user to restart the start command.
> 
> Steps to reproduce (in the TS example):
> 
>     1. Change API to an invalid name
> 
>     2. Change to the suggested name
> 
>     3. As it refreshes you can see the API has not been registered correctly

@davemooreuws can you confirm that this error does not occur on `84240a3214f43a0553ecc86297eeb53d0076200c`

I can reproduce it there which indicates that this is not a regression. We should raise a separate PR to fix this issue.

tjholm:
> @tjholm The errors show up nicely and it works well. However there is one scenario that forces the user to restart the start command.
> 
> Steps to reproduce (in the TS example):
> 
>     1. Change API to an invalid name
> 
>     2. Change to the suggested name
> 
>     3. As it refreshes you can see the API has not been registered correctly

Can you confirm this doesn't occur on `84240a3214f43a0553ecc86297eeb53d0076200c` (I can reproduce it there).

If it's a separate regression we should fix it in a separate PR to keep the diffs minimal.

pkg/cloud/resources/resources.go
```diff
@@ -134,9 +151,12 @@ func (l *LocalResourcesService) ClearServiceResources(serviceName string) {
 	l.state.ApiSecurityDefinitions.ClearRequestingService(serviceName)
 	l.state.SqlDatabases.ClearRequestingService(serviceName)
 	l.state.BatchJobs.ClearRequestingService(serviceName)
+	// TODO: lock
```
jyecusch: ?

pkg/cloud/schedules/schedules.go
```diff
@@ -177,12 +192,14 @@ func (l *LocalSchedulesService) Schedule(stream schedulespb.Schedules_ScheduleSe
...
 		if len(parts) != 2 {
-			return fmt.Errorf("invalid schedule rate: %s", t.Every.Rate)
+			l.errorLogger(serviceName, fmt.Errorf("invalid rate: %s for schedule %s", t.Every.Rate, scheduleName))
+			return nil
 		}
 
 		initialRate, err := strconv.Atoi(parts[0])
 		if err != nil {
-			return fmt.Errorf("invalid schedule rate, must start with an integer")
+			l.errorLogger(serviceName, fmt.Errorf("invalid rate: %s for schedule %s", t.Every.Rate, scheduleName))
```
jyecusch: the previos error description was more informative, I think we should also avoid any exact duplicate error description, since it makes it harder to narrow down where in the code they came from.

pkg/validation/name.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var ResourceName_Rule = &Rule{
+	name: "Invalid Name",
+	// TODO: Add docs link for rule when available
+	docsUrl: "",
+}
+
+// create a lower kebab case only regex
+var lowerKebabCase, _ = regexp.Compile("^[a-z0-9]+(-[a-z0-9]+)*$")
```
jyecusch:
```suggestion
var lowerKebabCaseRe, _ = regexp.Compile("^[a-z0-9]+(-[a-z0-9]+)*$")
```

pkg/validation/name.go
```diff
@@ -0,0 +1,41 @@
...
+
+func IsValidResourceName(name string) bool {
+	return lowerKebabCase.Match([]byte(name))
```
jyecusch:
```suggestion
	return lowerKebabCaseRe.Match([]byte(name))
```

pkg/cloud/resources/resources.go
tjholm: cleaned up

jyecusch (dismissed):

pkg/cloud/schedules/schedules.go
tjholm: have re-introduced description of error

HomelessDinosaur (approved):

davemooreuws (approved):

PR #814, opened 2024-10-31
feat: use latest provider version in stack new and stack up commands

### LINKED ISSUE #813: Automatic check for new official provider releases
### Suggestion
- fetch and store the latest provider version once every 24 hours
- use this version for any `nitric stack new` calls
- notify if stack version has an update in `nitric up` or `nitric version` calls

### Value
No need to manually update templates and bundle with CLI
More transparency when new releases are out

davemooreuws:
Closes #813 

- fetch and cache latest provider version
- use the latest version in `stack new`
- show message if a version is outdated during `stack up`

### DIFF
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/spf13/cobra"
 
 	"github.com/nitrictech/cli/pkg/paths"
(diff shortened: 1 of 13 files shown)


pkg/update/update.go
```diff
@@ -33,13 +33,55 @@ import (
...
+
+	var currentVersion string
+
+	providerParts := strings.SplitN(providerName, "@", 2)
+	if len(providerParts) == 2 {
+		providerName = providerParts[0]
+		currentVersion = providerParts[1]
+	}
+
+	if currentVersion < latestVersion {
```
HomelessDinosaur: Should the output only work if the provider format is correct? Or exit earlier. Also, the currentVersion < latestVersion comparison will fail when either values get to double digits because it's doing string comparisons. Might need to split it into parts and compare each part numerically.
i.e. "1.4.2" < "1.10.2" will be false

HomelessDinosaur (changes requested):

pkg/update/update.go
davemooreuws: I think at this stage the provider has been validated and downloaded. Yeah it probably should.

Good catch with the string values, this would also be the case with the CLI version check we already do.

HomelessDinosaur (approved):

tjholm (approved):

PR #815, opened 2024-11-01
fix: add podman support

jyecusch:
Fallback to Podman if Docker is unavailable.


jyecusch:
This still relies on https://github.com/nitrictech/examples/pull/82 to also make TS projects compatible with Podman, since they use a cache setting that's not supported by Podman `sharing=private`.

The choosen provider used in stack files also needs to support podman.

pkg/view/tui/dependency.go
```diff
@@ -17,84 +17,97 @@
...
+		c.message = "docker buildx is required to run this command. For installation instructions see: https://github.com/docker/buildx"
+		return err
+	}
+
+	return nil
+}
+
+func (c *ContainerToolDependency) Assist() string {
+	return c.message
 }
```
tjholm: Could dependency checks be a function with a new error type?

e.g.

```go
type DependencyMissingError struct {
	message string
}

func (d *DependencyMissingError) Error() string {
	return d.message
}

type Dependency = func() error

func ContainerToolDependency() error {
	_, err := DockerAvailable()
	if err != nil {
		_, podErr := PodmanAvailable()
		if podErr != nil {
			return &DependencyMissingError{
				message: "Docker or Podman is required, see https://docs.docker.com/engine/install/ for docker installation instructions",
			}
		} else {
			// Use podman
			return nil
		}
	}

	_, err = DockerBuildxAvailable()
	if err != nil {
		return &DependencyMissingError{
			message: "docker buildx is required to run this command. For installation instructions see: https://github.com/docker/buildx",
		}
	}

	return nil
}
```

jyecusch: I've updated it to make dependencies a function. I kept the error (why we think the dependency is missing) separate from the assist message (how we think it can be fixed). However, we're not really using the error right now. Maybe the design should change to just log that error with a debug log level so people can view it as needed, but we can throw it away and remove the second message from the struct.

tjholm (dismissed):

tjholm (dismissed):

davemooreuws (approved):

davemooreuws (changes requested):
I get this error when trying `nitric up` with no docker running:

```
 error  exit status 1
```

used to be

```
error  docker is required to run this command. For installation instructions see: https://docs.docker.com/engine/install/
```

davemooreuws (approved):

tjholm (approved):

PR #819, opened 2024-11-11
fix: correctly apply http proxy env and default

davemooreuws:
It was not possible to set NITRIC_HTTP_PROXY_PORT via the .env file; this update fixes that.


davemooreuws:
closed as these env variables should be blacklisted. Will open another PR.

PR #834, opened 2024-12-18
fix: exit 1 on provider errors in non-interactive mode

### LINKED ISSUE #90: Nitric Up GitHub action succeeds even though the build failed
### Issue

Nitric Up GitHub action still shows up as successful even though it appears the build failed.

### Steps

1. Create a Nitric project
2. Create a GitHub action with Nitric GH Action using up command to deploy
3. Have a bug in the Nitric project that will cause the build to fail
4. Observe an error message in the build logs, but that the GH Action is still successful:

image

### Expected

Expect the GH Action Workflow to fail

### Environment and setup information

- Nitric CLI Version: Nitric 1.43.0 
- Operating System: Ubuntu (GH Action Runner)
- Nitric dependencies and their versions, such as SDKs or Middleware: Nitric Python SDK 1.0.6
- Cloud providers you are deploying to and their version: nitric/aws@1.4.5

### Other info

Nil

davemooreuws:
fixes https://github.com/nitrictech/actions/issues/90

### DIFF
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -307,10 +307,13 @@ var stackUpdateCmd = &cobra.Command{
 
 		// Step 5b. Communicate with server to share progress of ...
 		if isNonInteractive() {
+			providerErrorDetected := false
+
 			fmt.Printf("Deploying %s stack with provider %s\n", stackConfig.Name, stackConfig.Provider)
 			go func() {
 				for update := range errorChan {
 					fmt.Printf("Error: %s\n", update)
+					providerErrorDetected = true
 				}
 			}()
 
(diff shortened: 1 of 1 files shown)


jyecusch (approved):

HomelessDinosaur (approved):

cmd/stack.go
```diff
@@ -307,10 +307,13 @@ var stackUpdateCmd = &cobra.Command{
 
 		// Step 5b. Communicate with server to share progress of ...
 		if isNonInteractive() {
+			providerErrorDetected := false
+
 			fmt.Printf("Deploying %s stack with provider %s\n", stackConfig.Name, stackConfig.Provider)
 			go func() {
 				for update := range errorChan {
 					fmt.Printf("Error: %s\n", update)
+					providerErrorDetected = true
```
tjholm, line 316: This looks good for now, because the error handling is in a go-routine there is a potential race here if the eventChan closes before any errors are received, which is unlikely to be a problem. We can fix this later by merging these channels into a single event stream or by using a select statement to make sure they're all recieved in the same routine.

PR #836, opened 2024-12-19
fix: handle migration path conflicts for identical database declarations

davemooreuws:
When multiple database declarations occur, check if the migrations path conficts and error with action.

### DIFF
--- a/pkg/collector/spec.go
+++ b/pkg/collector/spec.go
@@ -271,16 +271,40 @@ func GetMigrationImageBuildContexts(allServiceRequirements []*ServiceRequirement
 	return imageBuildContexts, nil
 }
 
+func checkConflictingMigrations(allDatabases []map[string]*resourcespb.SqlDatabaseResource, resource map[string]*resourcespb.SqlDatabaseResource) error {
+	for _, dbs := range allDatabases {
+		for databaseName, dbConfig := range resource {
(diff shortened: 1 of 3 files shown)


tjholm (dismissed):

raksiv (dismissed):

pkg/collector/spec.go
```diff
@@ -271,16 +271,40 @@ func GetMigrationImageBuildContexts(allServiceRequirements []*ServiceRequirement
 	return imageBuildContexts, nil
 }
 
+func checkConflictingMigrations(allDatabases []map[string]*resourcespb.SqlDatabaseResource, resource map[string]*resourcespb.SqlDatabaseResource) error {
+	for _, dbs := range allDatabases {
+		for databaseName, dbConfig := range resource {
+			if existing, exists := dbs[databaseName]; exists {
+				if existing.Migrations.GetMigrationsPath() != dbConfig.Migrations.GetMigrationsPath() {
```
raksiv, line 282: Is there a chance dbConfig.Migrations is nil ? might panic here - but not sure if it should error out or continue to process.

            if dbConfig.Migrations == nil {
                continue // or decide if this is an error
            }

raksiv (approved):

tjholm (approved):

PR #839, opened 2025-01-06
fix: update path matching regex to be stricter to ensure no history leakage

HomelessDinosaur:
Tested using the following paths:
```
/
/user
/user/{name}
/user/{name}/images
/user/{name}/images/{id}
```

Correctly associated each history path to belong to the correct API route. For reference: the regex looks like `^\/user\/[^/]+$` for a path to match for `/user/{name}`.

### DIFF
--- a/pkg/dashboard/frontend/src/components/apis/APIHistory.tsx
+++ b/pkg/dashboard/frontend/src/components/apis/APIHistory.tsx
@@ -27,7 +27,7 @@ const checkEquivalentPaths = (matcher: string, path: string): boolean => {
     return matcherBase === pathBase
   }
 
-  const regex = new RegExp(`^${matcherBase.replace(/{[^/]+}/, '[^/]+')}$`)
+  const regex = new RegExp(`^${matcherBase.replace(/{[^/]+}/g, '[^/]+')}$`)
 
   return regex.test(pathBase)
 }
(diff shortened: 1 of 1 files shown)


pkg/dashboard/frontend/src/components/apis/APIHistory.tsx
```diff
@@ -18,13 +18,18 @@ interface Props {
...
+  const [pathBase] = path.split('?')
+
+  // If both have query parameters, compare only the base paths
+  if (matcher.includes('?') && path.includes('?')) {
+    return matcherBase === pathBase
   }
 
-  const regex = matcher.replace(/{(.*)}/, '(.*)')
-  return path.match(regex) !== null
+  const regex = new RegExp(`^${matcherBase.replace(/{[^/]+}/, '[^/]+')}$`)
```
davemooreuws:
```suggestion
  const regex = new RegExp(`^${matcherBase.replace(/{[^/]+}/g, '[^/]+')}$`)
```

replace all placeholders not just first occurrence, checked with the /user/:name/images/:id path

davemooreuws (changes requested):

davemooreuws (approved):

tjholm (approved):

PR #856, opened 2025-03-07
feat: add website command

### LINKED ISSUE #855: Add websites to projects via CLI
### Suggestion
Add the ability to add websites during nitric new and / or via another command.

### Value
Reduces maintenance burden of having a large matrix of cross language projects in the starter list

davemooreuws:
closes #855


tjholm (dismissed):

pkg/view/tui/commands/website/new.go
```diff
@@ -0,0 +1,577 @@
...
+		} else {
+			step = StepTool
+		}
+	}
+
+	if args.WebsitePath != "" {
+		// check if the path is already in use
+		if lo.Contains(existingPaths, args.WebsitePath) {
+			return Model{}, fmt.Errorf("path %s is already in use", args.WebsitePath)
+		}
```
jyecusch:
```suggestion
		pathInUse := lo.Contains(existingPaths, args.WebsitePath)
		if pathInUse {
			return Model{}, fmt.Errorf("path %s is already in use", args.WebsitePath)
		}
```

I think code should be used over comments to explain what's going on, it may not even be needed here, the `if` statement was already pretty clear.

Please replace/remove the code comments through, anywhere they explain *what* the code is doing.

jyecusch (changes requested):

jyecusch (approved):

tjholm (approved):

